PR #4, opened 2022-08-17
Added support for rke2 kubeadm and containerd packages

venkatnsrinivasan:
Added support for the following packages
- Containerd 
- kubeadm (installs kubeadm/kubelet/kubectl as per CNCF install instructions )
- RKE2 (currently rke2 does not support ARM so the arm flag has no effect)


mudler (approved):
Looks good here, thanks!

PR #114, opened 2023-02-14
Add guards for templates

Itxaka:
If you try to build locally some issues will arise due the templating trying to access non existing values.

This patch guards those failures by first checking the the var exists before trying to get the value, avoiding panics.

Signed-off-by: Itxaka <itxaka.garcia@spectrocloud.com>


mudler (approved):
no version bumps huh.. ;D

PR #118, opened 2023-02-15
Added a legacy package for immutable-rootfs module which uses an older version of lsblk

venkatnsrinivasan:
Added a legacy package for immutable-rootfs module which uses an older version of lsblk

### DIFF
--- a/packages/system/dracut/immutable-rootfs-legacy-compat/30cos-immutable-rootfs/cos-loop-img.sh
+++ b/packages/system/dracut/immutable-rootfs-legacy-compat/30cos-immutable-rootfs/cos-loop-img.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+function doLoopMount {
+    local partdev
+    local partname
+    local dev
+
+    # Iterate over current device labels
+     for partdev in $(lsblk -nlp | awk '$0 ~ /part/ {print $1}'); do
+        partname=$(basename "${partdev}")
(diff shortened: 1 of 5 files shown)


packages/system/dracut/immutable-rootfs-legacy-compat/30cos-immutable-rootfs/cos-loop-img.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+function doLoopMount {
+    local partdev
+    local partname
+    local dev
+
+    # Iterate over current device labels
+     for partdev in $(lsblk -nlp | awk '$0 ~ /part/ {print $1}'); do
```
Itxaka, line 9: I guess the change is in the output of lsblk?

Maybe we can work around by checking before the output or version or some switches to identify the correct way to call lsblk?

That would mean compatibility with both versions

venkatnsrinivasan: I had a conversation with @mudler  and we agreed to keep this a separate package as its temporary and eventually the use of the lsblk based config will be deprecated.

mudler: yep, no need to be backward compatible as with immucore this stuff won't be required anymore

mudler (approved):

PR #135, opened 2023-03-06
Add shim-signed package

Itxaka:
Provides the signed artifacts shim and grub for secure boot


Itxaka:
Not sure how I feel about this. It maybe better to drop this and shim and just include all artifacts in the grub2-efi package instead?

PR #136, opened 2023-03-06
Bundle signed grub+shim in grub2-efi package

Itxaka:


Itxaka:
probably superseedes https://github.com/kairos-io/packages/pull/135 and makes system/shim deprecated

mauromorales (approved):
yup I think it's better both together to see the branching caused by the arch :+1:

Itxaka:
This also provides the shim for both arches, which is not the case in the shim package, it only provides it for x86.
Also provides both the signed grub and shim which are required for secureboot.

also matches with https://github.com/kairos-io/elemental-cli/pull/44 which uses those paths to use the grub artifacts during the install

packages/system/grub2-efi/build.yaml
```diff
@@ -1,13 +1,22 @@
...
+steps:
+- zypper in grub2-arm64-efi shim-susesigned
+includes:
+- usr/share/grub2/arm64-efi/
+- usr/share/efi/.*/*.efi
+- usr/share/efi/.*/*.der
+  {{else}}
 steps:
-- mkdir -p /data
-- wget  https://download.opensuse.org/repositories/isv:/Rancher:/Elemental/15.4/noarch/grub2-{{.Values.arch}}-efi-{{.Values.pversion}}.noarch.rpm -O /package.rpm
```
Itxaka, line 8: this doesnt seem to exist anymore? Plus now elemental uses upstream grub IIRC.

mudler (approved):

PR #151, opened 2023-03-13
Fix system/grub2-efi package

Itxaka:
Files in the shim package are already signed AND thye provide more efi tools, like the mokmanager or a fallback efi binary


packages/system/grub2-efi/build.yaml
```diff
@@ -6,14 +6,14 @@ prelude:
 {{ if .Values.arch }}
   {{ if eq .Values.arch "arm64" }}
 steps:
-- zypper in -y grub2-arm64-efi shim-susesigned
```
mudler, line 9: indeed.. I didn't dared to ask about it. my bad :monkey:

PR #160, opened 2023-03-24
:arrow_up: Bump elemental

Itxaka:
Some idiot forgot to bump the hash so the elemental package only changed the version number, content was pointing to the same hash...ains.


mauromorales (approved):
and I think it was my fault to review and not say anything :see_no_evil: ... lgtm

Itxaka:
> and I think it was my fault to review and not say anything see_no_evil ... lgtm

I also was expecting to work like other packages, bump version picks new tag from git.... :D

PR #175, opened 2023-04-11
added amt provider package

paynejacob:


packages/utils/provider-amt/build.yaml
```diff
@@ -0,0 +1,18 @@
...
+    source: /usr/local/lib/librpc.so
+    destination: /usr/local/lib/librpc.so
+  - image: "{{ .Values.labels.repository }}:v{{ .Values.version }}"
+    source: /usr/local/include/librpc.h
+    destination: /usr/local/include/librpc.h
+  - image: "{{ .Values.labels.repository }}:v{{ .Values.version }}"
+    source: /system/providers/agent-provider-amt
+    destination: /usr/local/system/providers/agent-provider-amt
+steps:
+  - chmod +x /usr/local/system/providers/agent-provider-amt
```
mudler, line 14: @paynejacob I thought we would use it as a bundle and not as a provider?

Currently the agent doesn't support multiple bundles, and this means you can't use amt with others - or I am missing something?

packages/utils/provider-amt/build.yaml
```diff
@@ -0,0 +1,18 @@
...
+includes:
+  - /usr/local/lib/librpc.so
```
mudler, line 16: I'm wondering if this is really required? Cant we embed the libs statically?

paynejacob: ill try a build with c go off and see if we can skip it.

packages/utils/provider-amt/build.yaml
paynejacob: This bundle will install amt as a provider.  The idea being if you are using amt to manage your hardware you would use the os builder to create an iso with the amt bundle installed.  AMT only needs to be configured once so it makes sense to run it on install.  Are you saying we can only have one bundle in the install config or one bundle per system?

mudler: yes, at the moment we can have only one bundle only - they don't play nicely together yet

mudler: see https://github.com/kairos-io/kairos/blob/afc1a13ab8053f76f0a0d1c54ebabc4686253abe/internal/agent/install.go#L115 - we should do some merging strategy in all the places were we call providers in order to support multiple providers

mudler: what about making it a bundle? that should work as well and can also bring binaries with it (thus, you need to place them for instance in a persistent location, as `/usr/local/bin`)

paynejacob: I see we want it to be a systemd service, I can make that change.

packages/utils/provider-amt/build.yaml
paynejacob: It looks like this library cannot be embedded but I was able to link it to a static file so we don't need to clear the ld cache.

Itxaka, line 16: This seems to be becuase on our provider we link to C libraries directly. Why is that @paynejacob ? I saw that intel provides a pure go client for atm at https://github.com/open-amt-cloud-toolkit/rpc-go which directly uses the rpc stuff in there with no linking. 

Maybe we can use that so we also have a pure go lib for this?

*Disclaimer*: Is like the first time I heard about amt so no idea of the drawback of using go or how it even works, so maybe my comment is wrong!

paynejacob: That was my first approach 😄  to just use their library but all the functionality we needed was in the internal package so it wasn't accessible.  The alternative would be to make a subprocess call to their rpc binary but imo subprocess calls are fragile.  I noticed in their docs you could build the library as a c library and it allowed for making rpc calls directly.  This was the cleanest solution I could find.  I am open to revisiting any of these, given we are pivoting to a bundle I think the easiest thing would be to tell our binary to search it's current directory for the lib and include the library in the bundle.  We could also use the subprocess call to their binary but I think that is about the same complexity.

Itxaka: ahhh...go and its stupid internal dir ¬_¬

Maybe we can open an issue on the repo so they can extract that to a different dir so its available as a lib for other projects? Or, as its apache, we can just bring the whole internal dir into our repo directly as long as we keep the apache license :D

paynejacob: Ill open that issue, I agree just importing their code would be ideal.  For now I have it working as a bundle with the library and binary.  The binary checks for the library in it's current directory so it will be packageable as a bundle.  I think it will be easier than trying to maintain a fork for now.

packages/utils/provider-amt/build.yaml
mudler: right indeed, no need even to copy anything, if it's a one-shot command, it can just register and be done with it!

PR #182, opened 2023-04-13
Provide bios files in arm64 livecd packages

Itxaka:
elemental-cli still builds hybrid isos for all arches, so it tries to find some files in the arm64 packages which are not there.

This patch makes it so the arm64 package provides those files like the x86 package until we change the iso builder to not generate hybrid isos for arm64


mudler (approved):
looks good over here!

mauromorales (approved):
lgtm

PR #226, opened 2023-05-12
Track elemental-cli by version not by hash

Itxaka:
we were tracking the versions via hash which is not nice.

Track by version like the rest of the packages


mauromorales (approved):
lgtm

PR #236, opened 2023-05-19
Bump distro-initrd packages

Itxaka:
Changed the build to depend on kairos-agent but did not bump the packages


Itxaka:
requires more changes!

PR #237, opened 2023-05-19
Fix alpine initrd

Itxaka:
Was missing the immucore stuff, add the kairos-agent, drop kcrypt


mauromorales (approved):
does this mean we can stop relying on the opensuse or ubuntu kernels?

Itxaka:
> does this mean we can stop relying on the opensuse or ubuntu kernels?

Niet. It means that alpine flavors were not using immucore properly :laughing:

Itxaka:
cc @mudler as the expert on all things luet, would love your view here.

I could not bump initrd by itself because it required a base-image bump to pick the correct arch. Also had to add the amd64 arch values, otherwise the initrd would get an empty arch on x86 and not install the proper packages.

.github/workflows/build.yaml
```diff
@@ -17,16 +17,12 @@ jobs:
...
-      run: |
-        sudo rm -rf /usr/local/lib/android # will release about 10 GB if you don't need Android
-        sudo rm -rf /usr/share/dotnet # will release about 20GB if you don't need .NET
     - name: Download meta 🔧
       uses: luet-lab/luet-github-action@master
       with:
         FINAL_REPO: quay.io/kairos/packages
         REPOSITORY_TYPE: docker
-        downloadAllMeta: true
+        downloadAllMeta: false
```
mudler, line 25: that goes back to: https://github.com/Luet-lab/luet-github-action/blob/d8cbcf3057535de7ea33896b8ed44fc7b122d6b0/main.go#L371

mudler:
looks good and dandy here, I think the CI change is an old relic laying around, good catch @Itxaka !

PR #243, opened 2023-05-23
Run on self-hosted to unblock

Itxaka:


Itxaka:
cc @mudler merging this so it cvan build all packages in my arm worker as they are way too many and get out of space :/

Will revert once its pushed to the final repos and it doesnt need that much space

PR #260, opened 2023-05-31
Wait for cdrom in alpine initrd

Itxaka:
Sometimes cdrom can take a bit to appear as we load the modules and run udev.

Lets wait a bit fir it to appear before continuing


Itxaka:
last one I promise

PR #261, opened 2023-05-31
Bump grub to 2.6 and set terminal_output for arm images

mauromorales:
relates to https://github.com/kairos-io/kairos/issues/507

### DIFF
--- a/packages/static/grub-config/files/grub.cfg
+++ b/packages/static/grub-config/files/grub.cfg
@@ -54,6 +54,7 @@ insmod all_video
 insmod gfxterm
 insmod loopback
 insmod squash4
+terminal_output gfxterm
 
 menuentry "${display_name}" --id cos {
   search --no-floppy --label --set=root COS_STATE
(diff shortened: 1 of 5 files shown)


packages/static/grub-config/files/grub.cfg
```diff
@@ -54,6 +54,7 @@ insmod all_video
 insmod gfxterm
 insmod loopback
 insmod squash4
+terminal_output gfxterm
```
jimmykarily, line 57: I wonder if it's possible that we break the serial console with this setting (https://docs.fedoraproject.org/en-US/quick-docs/bootloading-with-grub2/#using-old-graphics-modes). I don't understand the implications of changing this value.

mauromorales: good question, I'll test it out to be sure it doesn't

Itxaka (approved):

packages/static/grub-config/files/grub.cfg
mauromorales: very nice catch @jimmykarily it was failing indeed, but I managed to find a solution, will merge and try everything with the osbuilder

PR #264, opened 2023-06-01
Use autobump for alpine kernels+initrd

Itxaka:
So we dont lag behind. This needs manual change when new alpine versions are released unfortunately


packages/alpine/collection.yaml
```diff
@@ -3,7 +3,23 @@ packages:
     category: "distro-kernel"
     version: "0.1.4"
     description: "Provides kernel and custom initrd for alpine"
+    labels:
+      autobump.strategy: "custom"
+      autobump.string_replace: '{ "prefix": "" }'
+      autobump.prefix: "prefix"
+      autobump.hook: |
+        curl -s https://raw.githubusercontent.com/alpinelinux/aports/3.18-stable/main/linux-lts/APKBUILD | grep "pkgver=" | awk -F"=" '{print $2}'
```
Itxaka, line 11: @mudler do you know if I can make the `3.18` in here a key instead? So in the build we use the image alpine:VERSION and here it bumps only for that same VERSION, so we can bump it in just one place and then it autobumps the kernel for that version forever.

Otherwise, this needs a manual bump to 3.19 when its released which means we will probably forget about it.

PR #280, opened 2023-06-13
Build fips packages with go 1.19.10

Itxaka:
Go 1.19.10 is the verified fips version from our side so we need to build those packages with that specific version.

This patch adds a go tollchain fips specific that is pinned to the know verified fips go version and that its used to build the fips verified go packages

Fixes: https://github.com/kairos-io/kairos/issues/1498

### DIFF
--- a/packages/system/kairos-agent/build.yaml
+++ b/packages/system/kairos-agent/build.yaml
@@ -1,8 +1,18 @@
+{{if eq .Values.category "fips" }}
+requires:
+  - name: "toolchain-go-ubuntu"
+    category: "fips"
+    version: ">=0"
+{{else}}
 requires:
   - name: "toolchain-go-ubuntu"
     category: "development"
     version: ">=0"
-
+{{end}}
+prelude:
+  - apt-get update && apt-get install -y npm gcc
+  - mkdir /go/src/github.com/${GITHUB_ORG}/ -p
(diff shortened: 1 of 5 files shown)


packages/system/kairos-agent/build.yaml
```diff
@@ -1,7 +1,14 @@
+{{if eq .Values.category "fips" }}
+requires:
+  - name: "toolchain-go-ubuntu"
+    category: "fips"
+    version: ">=0"
+  {{else}}
```
mauromorales, line 6: this is nitpicking but could you have the "else" and "end" match the indentation of the "if"?

packages/system/kcrypt/build.yaml
```diff
@@ -1,7 +1,15 @@
+{{if eq .Values.category "fips" }}
+requires:
+  - name: "toolchain-go-ubuntu"
+    category: "fips"
```
mauromorales, line 4: is there no way to inherit the category?

mauromorales (approved):
lgtm

packages/system/kairos-agent/build.yaml
Itxaka: crap. I swear its the editor auto changing this on save

packages/system/kcrypt/build.yaml
Itxaka: oh I see, good point I can just use `{{.Values.category}}` here! :facepalm:

PR #302, opened 2023-06-20
Update cos-setup-network.service to look like the other services

Itxaka:
Currently the network service is not kept as alive like the other and has some other configs in there which dont make much sense.

This patch just makes it equal to the rest of the other oneshot services


mauromorales (approved):
lgtm

Itxaka:
closing in favour of https://github.com/kairos-io/packages/pull/303 which is more complete

PR #306, opened 2023-06-21
Skip the fips check under arm64

Itxaka:
For some reason the output of the strings do not work properly under arm64. Skip it for now until we implement it the same way as immucore does


Itxaka:
only happens on master when it tries to build the missing packages

PR #334, opened 2023-07-18
Add provider-kairos package

jimmykarily:
as part of this: https://github.com/kairos-io/kairos/issues/1609


jimmykarily:
We already build a provider-kairos package here: https://github.com/kairos-io/packages/tree/main/packages/bundles/provider-kairos No fips but should be ok for now. Closing this.

PR #343, opened 2023-07-24
Fix default grub entry name to kairos

Itxaka:
As we dont ship the elemental config by default, this should fallback into a sane value so we dont have to have extra configs lying around.


mauromorales (approved):
😎

PR #351, opened 2023-07-27
Build firmwares for all arches

Itxaka:
So any luet installing from any arch can access the same packages


mudler (approved):
Looking good!

PR #353, opened 2023-07-31
Fix setting agent version in ldflags

Itxaka:
Name changed of the module


jimmykarily:
Does this fix an existing Github Issue? Good catch! I wonder how this went unnoticed for so long.

jimmykarily (approved):

PR #355, opened 2023-08-01
Add older version of kairos-agent v2.1.12

mauromorales:
This package is only needed temporarily so the image gets built with it


Itxaka (approved):
nice

mudler (approved):
Looks good!

PR #359, opened 2023-08-03
Test for squashfs recovery

Itxaka:
Instead of bindly doing a search which results in a error if recovery is not squash, just do a test to see if the file exit and set the parameters based on that

This should stop that ugly error message on recovery grub menu selection


Itxaka:
image

mauromorales (approved):
nice

Itxaka:
lets go and pray that I dont break the whole boot process lol

PR #364, opened 2023-08-04
Full cleanup of CI workers before starting

Itxaka:


Itxaka:
canceled build as this affects main CI only.

PR #376, opened 2023-08-16
Use diferent base images for livecd grub artifacts

Itxaka:
Looks like only the tumbleweed shim is properly signed to allow secureboot

And we are already using the tumbleweed artifacts in the normal grub2-efi packages so it makes sense to have them synced


packages/livecd/grub2/config/grub_live_bios.cfg
```diff
@@ -0,0 +1,70 @@
+
+search --file --set=root /boot/kernel
+set default=0
+set timeout=10
+set timeout_style=menu
+set linux=linux
+set initrd=initrd
+if [ "${grub_cpu}" = "x86_64" -o "${grub_cpu}" = "i386" -o "${grub_cpu}" = "arm64" ];then
+    if [ "${grub_platform}" = "efi" ]; then
+        if [ "${grub_cpu}" != "arm64" ]; then
```
mauromorales, line 10: if you remove the arm64 on line 8, you don't need to have this one

Itxaka: This is a copy of the grub.cfg from kairos directly. Im not touching that just in case as Im already changing a few things here and dont want to break it further lol

Itxaka: I dont even think this works, because line 16 should trigger on efi systems and ask to press t, but I have not seen that anywhere happening xD

mauromorales (approved):
lgtm

PR #377, opened 2023-08-16
Use tumbleweed artifacts for efi

Itxaka:
Looks like only the tumbleweed shim is properly signed to allow
secureboot


Itxaka:
This makes EFI work back again and secureboot as well.

mauromorales (approved):

PR #381, opened 2023-08-17
Create kairos-overlay-files package

jimmykarily:
as part of: https://github.com/kairos-io/kairos/issues/1726

The package contains whatever files are left in this branch: https://github.com/kairos-io/kairos/tree/1726-move-logic-to-overlays/overlay

(ignore files-iso which is going to be removed here: https://github.com/kairos-io/kairos/pull/1727)

After this package is published, Earthly in kairos will just install these files in the framework-luet target.


Itxaka:
Is this just the files copied over no? No changes to them or anything?

packages/bundles/kairos-overlay-files/collection.yaml
```diff
@@ -0,0 +1,8 @@
+packages:
+  - name: "kairos-overlay-files"
+    category: "bundles"
+    version: "2.4.0"
```
Itxaka, line 4: how we ganno manage this version? I guess independent from the kairos repo rigth? Its just a match for now

packages/bundles/kairos-overlay-files/collection.yaml
```diff
@@ -0,0 +1,8 @@
...
+    labels:
+      github.repo: "kairos"
+      github.owner: "kairos-io"
+      autobump.strategy: "release"
```
Itxaka, line 8: ah, this wont work rigth? It needs kairos to be bumped and then this will be bumped but that means when we bump kairos to 2.5.0 this will get bumped to the same version BUT the package installed on kairos 2.5.0 wont be that one....

I would not let it autobump, especially with no changes.

packages/bundles/kairos-overlay-files/collection.yaml
Itxaka: just saw below....I would set this to 1.0.0 and leave it unconnected to the kairos version for now, as bumps to this package do not neccesarily translate to the kairos version no?

Itxaka (dismissed):
looks good but looking for clarification in the versioning scheme

EDIT: This was supposed to be a comment, sorry

packages/bundles/kairos-overlay-files/collection.yaml
jimmykarily: yes, I wasn't sure either. My plan was to start from what kairos is now and let it diverge. But I think your suggestion is better. Let's start with 1.0.0 to make it clear that they have no relation.

packages/bundles/kairos-overlay-files/collection.yaml
jimmykarily: You are right. I had no idea what this did. I just copied from the [other package](https://github.com/kairos-io/packages/tree/main/packages/bundles/kairos-default-configs). Btw that package is just a subset of this new one. I wonder if that is being used at all. Maybe we should remove it?

jimmykarily:
> Is this just the files copied over no? No changes to them or anything?

yes, zero changes

mauromorales (approved):

packages/bundles/kairos-overlay-files/collection.yaml
Itxaka: I dont hink tha tits used anywhere but at least the bumping works...

Itxaka (approved):
awesome!

PR #387, opened 2023-08-21
fix: do not overlay mount persistent in recovery mode

mudler:
Potential fix for: https://github.com/kairos-io/kairos/issues/1723

In orin, we unify in a single overlay `/usr/local` from the image and COS_PERSISTENT. This creates a new mountpoint that gets in the way of running the recovery. Since this is only needed (the transparent mount) when running the workload - this PR disables the overlay mounting during `reset`.

I'm not able to test it for now as it seems upgrading the recovery is broken (https://github.com/kairos-io/kairos/issues/1751 ) - and I have to re-flash the board again as other attempts seems to have nuked my drive.


mauromorales (approved):
lgtm

PR #392, opened 2023-08-22
Remove bootargs.cfg from static package

jimmykarily:
because it gets applied and replaces the ones here:

https://github.com/kairos-io/kairos/pull/1756/files


jimmykarily:
> I think this would break the osbuilder tools https://github.com/kairos-io/osbuilder/blob/master/tools-image/Dockerfile#L52

we discussed this on Slack, it should break something but let's see

mauromorales (approved):

PR #398, opened 2023-08-28
Automatically expand persistent partition on RPI 4

jimmykarily:
Fixes: https://github.com/kairos-io/kairos/issues/1762
and: https://github.com/kairos-io/kairos/issues/1448

### DIFF
--- a/packages/bundles/kairos-overlay-files/collection.yaml
+++ b/packages/bundles/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "bundles"
-    version: "1.0.4"
+    version: "1.0.5"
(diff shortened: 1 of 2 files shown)


jimmykarily:
If we like it, we'll need to also bump the package version so that we can release it.

mauromorales:
I guess it will need a bump of a version somewhere

I wonder if we really need this because the partition is properly expanded sometimes ... what is the easiest way to test this before we release package?

jimmykarily:
> I guess it will need a bump of a version somewhere
> 
> I wonder if we really need this because the partition is properly expanded sometimes ... what is the easiest way to test this before we release package?

I just added that to my cloud-config as described here: https://kairos.io/docs/installation/raspberry/#configure-your-node

To test it as a luet package is more complicated. You'd have to create a luet repository manually (a docker image) and push it somewhere, then refer to it in the framework-profile.

jimmykarily:
I bumped the package version

PR #405, opened 2023-09-04
Grow fs as part of the default config for rpi4

Itxaka:
Fixes https://github.com/kairos-io/kairos/issues/1783

### DIFF
--- a/packages/bundles/kairos-overlay-files/files/system/oem/11_persistent.yaml
+++ b/packages/bundles/kairos-overlay-files/files/system/oem/11_persistent.yaml
@@ -11,3 +11,6 @@ stages:
         expand_partition:
           # Size 0 is required to specify all remaining space
           size: 0
+      commands:
+        # grow filesystem
+        - resize2fs /dev/disk/by-label/COS_PERSISTENT
--- a/packages/bundles/kairos-overlay-files/collection.yaml
+++ b/packages/bundles/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "bundles"
-    version: "1.0.5"
+    version: "1.0.6"


packages/bundles/kairos-overlay-files/files/system/oem/11_persistent.yaml
```diff
@@ -11,3 +11,6 @@ stages:
         expand_partition:
           # Size 0 is required to specify all remaining space
           size: 0
+      commands:
+        # grow filesystem
+        - resize2fs /dev/disk/by-label/COS_PERSISTENT
```
mudler, line 16: shouldn't we call also growpart first? `resize2fs` just enlarges the fs IIRC

packages/bundles/kairos-overlay-files/files/system/oem/11_persistent.yaml
```diff
@@ -11,3 +11,6 @@ stages:
         expand_partition:
```
mudler, line 11: expand_partition works only on GPT, if we call growpart it would work as well for RPI3

mudler:
I am a bit confused - rpi4 is GPT and the expand_partition should work with GPT partitioned disk right? isn't this necessary only for RPI3? and needs a combination of growpart + fs resizing

packages/bundles/kairos-overlay-files/files/system/oem/11_persistent.yaml
Itxaka: that what we want no? The partition is already resized but the FS is not, so its only taking a small part of the partition. We only want to resize the fs, the expansion is done by yip. In fact this is a workaround, the real fix should go into the layout plugin to call the fs expansion after the partition recreation to not leave this like this.

Both growpart and resize2fs work the same, they resize the FS. Not sure if growpart is already installed in our images but resize2fs is, as its part of e2fsprogs

packages/bundles/kairos-overlay-files/files/system/oem/11_persistent.yaml
Itxaka: ?

This is the yip plugin which works only on GPT. For MBR there is nothing that can expand or create partitions currently.

packages/bundles/kairos-overlay-files/files/system/oem/11_persistent.yaml
mudler: growpart resizes a partition to my knowledge, and NOT the filesystem. resize2fs resizes the FS, NOT the partition (they are doing two different things).

I think to make things more simpler, if we need many workaround we can just drop the usage of the cloud config here and just use growpart+resize2fs which works for both MBR and GPT. 

There is one subtle issue with this too which is this: https://github.com/kairos-io/kairos-agent/blob/307eabbb36c9c6969681f4a3852e53d4dd572a2e/pkg/partitioner/disk.go#L359 I've been facing this myself - and the result was unpleasant. If we try to extend the partitions regardless on each boot, there was a change that the partition was completely nuked between standard reboots

packages/bundles/kairos-overlay-files/files/system/oem/11_persistent.yaml
mudler: nope but just by using growpart+resize2fs would do the trick for both MBR and GPT

PR #412, opened 2023-09-08
Split k3s package in 2 (one for openrc, one for systemd)

jimmykarily:
because having all service files in places, confuses systemd because of the sysv compatibility feature.

Part of: https://github.com/kairos-io/kairos/issues/1797

### DIFF
--- a/packages/k8s/k3s-openrc/build.yaml
+++ b/packages/k8s/k3s-openrc/build.yaml
@@ -1,0 +1,32 @@
+requires:
+- name: "toolchain-go"
+  category: "development"
+  version: ">=0"
+env:
+  - INSTALL_K3S_BIN_DIR="/usr/bin"
+  - INSTALL_K3S_SELINUX_WARN=true
+  - INSTALL_K3S_SKIP_START="true"
+  - INSTALL_K3S_SKIP_ENABLE="true"
+  - INSTALL_K3S_SKIP_SELINUX_RPM="true"
+  # Remove any possible "+N" from the end of the version. We only bump that to
+  # avoid overwritting existing images when we want the package to be rebuilt.
+  - INSTALL_K3S_VERSION=v{{ regexReplaceAll "\\+\\d+$" .Values.version "" }}+k3s1
+  {{$arch:=(default "amd64" .Values.arch)}}
+  {{ if eq $arch "arm" }}
+  - ARCH=arm64
+  {{ else }}
+  - ARCH={{ $arch }}
+  {{ end }}
+steps:
+  - curl -sfL https://get.k3s.io > installer.sh
+  # Let the installer script install service files for openrc:
+  # https://github.com/k3s-io/k3s/blob/36645e7311e9bdbbf2adb79ecd8bd68556bc86f6/install.sh#L114-L122
+  - touch /sbin/openrc-run && chmod +x /sbin/openrc-run
+  - bash installer.sh
+  - rm -rf installer.sh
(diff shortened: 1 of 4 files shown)


mudler (approved):

packages/k8s/k3s-openrc/collection.yaml
```diff
@@ -1,7 +1,7 @@
 packages:
   - name: k3s
     category: k8s
-    version: "1.25.11+1"
+    version: "1.25.11"
```
Itxaka, line 4: This wont work no? the package with +1 will be the latest in luet eyes I think @mudler can confirm?

packages/k8s/k3s-systemd/collection.yaml
```diff
@@ -0,0 +1,36 @@
+packages:
+  - name: k3s
```
Itxaka, line 2: name is not changed, there will be two different packages with the same name and category, this need to be changed to k3s-systemd or k3s-openrc

packages/k8s/k3s-openrc/build.yaml
```diff
@@ -0,0 +1,32 @@
+requires:
```
Itxaka, line 1: you dont need another build file, you can use templating to generate one package or the other in the same build file!

packages/k8s/k3s-systemd/build.yaml
```diff
@@ -19,27 +19,15 @@ env:
   {{ end }}
 steps:
   - curl -sfL https://get.k3s.io > installer.sh
-  # Let the installer script install service files both for openrc and systemd:
+  # Let the installer script install service files for systemd:
   # https://github.com/k3s-io/k3s/blob/36645e7311e9bdbbf2adb79ecd8bd68556bc86f6/install.sh#L114-L122
   - touch /bin/systemctl && chmod +x /bin/systemctl
   - mkdir -p /etc/systemd/system/
   - bash installer.sh
   - bash installer.sh agent
```
Itxaka, line 27: you can reuse this and use templating to execute the below commands only if the package is called k3s-openrc or k3s-systemd

Itxaka (changes requested):
Al least the package split in naming should be done and the version rev bump, otherwise its not gonna work

packages/k8s/k3s-openrc/collection.yaml
jimmykarily: We split the package into 2 new ones. So for luet, this is the first time it sees this (new) package.

But...we forgot to change the name of the package. Doing that now.

packages/k8s/k3s-systemd/collection.yaml
jimmykarily: yes, saw that now. Fixed.

packages/k8s/k3s-openrc/build.yaml
jimmykarily: They are quite different on the way they fake openrc/systemd, the files they copy and such. We would remove duplication but introduce templating complexity. I'd rather keep them separate for now. What do you think?

Itxaka: Good to me!

jimmykarily: Nevermind, we tried the templating and it turns out it doesn't look too complicated. We are going with your suggestion.

mudler (approved):

Itxaka (approved):

PR #416, opened 2023-09-11
Copy the symlink of the openrc enabled service

jimmykarily:
because k3s is not enabled in alpine otherwise


jimmykarily:
provider-kairos should have enabled the service, the bug is elsewhere. Closing this.

PR #417, opened 2023-09-11
Enable openrc k3s

jimmykarily:


packages/k8s/k3s/build.yaml
```diff
@@ -37,6 +37,7 @@ includes:
 {{ if eq .Values.name "k3s-openrc" }}
 - ^/etc/init.d/$
 - ^/etc/init.d/k3s.*
+- ^/etc/rancher/k3s$
```
Itxaka, line 40: we m,ay need to include the /etc/rancher dir first then this one

Itxaka (approved):

PR #441, opened 2023-09-15
Drop unused packages

Itxaka:
Also:
 - merges systemd network config files with systemd serviuce package files as they always go together
 - changes category the kairos overlay files to static as it made not much sense to be in bundles and could lead to confusion with our kairos "bundles"

After merge, this requires:

 - Dropping the systemd-networkd package from kairos
 - Changing the overlay package to use the proepr category on kairos


jimmykarily:
How do we know if something is not used or not? Is it enough to check if there are references to those in our other repos? Which ones? Do we care if other people/forks/projects rely on these packages in any way?

Itxaka:
> How do we know if something is not used or not? Is it enough to check if there are references to those in our other repos? Which ones? Do we care if other people/forks/projects rely on these packages in any way?

I checkd our packages and those are not used anywhere.

And the official way of building a kairos derivative is either via using one of our published imagesx as base or the framework. Knowing that only the framework is consuming that, I would say this is safe enough to drop. 


And if not we will get complaints very soon and can restore whatever package is being used. Not only that but if they are pinnned to versioned repos, this wont break anything until they bump the repo to a newer version and then starts to fail, so again, should eb safe enough to not break anything current using this repos, if someone even is doing it. While we remove 1500 lines of useless code....

jimmykarily (approved):

PR #450, opened 2023-09-22
Split uki config and bump package

Itxaka:
Provides uki rootfs config


Itxaka:
Doesnt touch normal systems, so merging ti directly

PR #473, opened 2023-10-09
Drop gfxterm

Itxaka:
Unless we really configure it for the lower common denominator is proving to be a pita.
Duplicated terminal outputs, wrong videomodes, and now out of memory when used in 4k devices.

This just drops the gfxterminal for now to stay with a text terminal for grub

### DIFF
--- a/packages/static/grub-config/definition.yaml
+++ b/packages/static/grub-config/definition.yaml
@@ -1,3 +1,3 @@
 name: "grub-config"
 category: "static"
-version: "0.7"
+version: "0.8"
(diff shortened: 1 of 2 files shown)


mauromorales (approved):
LGTM, would be good to document somewhere in case someone want's to use the serial console so they don't have to figure out from scratch

Itxaka:
> LGTM, would be good to document somewhere in case someone want's to use the serial console so they don't have to figure out from scratch

the Idea here is once merged built locally and try in the rpi+serial to see if the issue was this, as I remember it breing broken due to some changes that I did and that I reverted back them, so maybe its not needed for the gfxterminal. I mean, text should be the most compatible one, working on everything...

PR #474, opened 2023-10-09
Add missing files to static configs

Itxaka:
Moving them out of the kairos repo to be used in the framework

Part of: https://github.com/kairos-io/kairos/issues/511

### DIFF
--- a/packages/static/kairos-overlay-files/collection.yaml
+++ b/packages/static/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.0.14"
+    version: "1.1.0"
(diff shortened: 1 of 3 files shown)


jimmykarily:
There was some logic on when to copy the bootargs: https://github.com/kairos-io/kairos/blob/be7961bc8086838c378839364a9f0e77861ecdbd/Earthfile#L277

(we didn't copy that when it was there already). Would that break something by replacing an existing bootargs file in some cases?

jimmykarily (approved):

Itxaka:
> There was some logic on when to copy the bootargs: [kairos-io/kairos@`be7961b`/Earthfile#L277](https://github.com/kairos-io/kairos/blob/be7961bc8086838c378839364a9f0e77861ecdbd/Earthfile#L277)
> 
> (we didn't copy that when it was there already). Would that break something by replacing an existing bootargs file in some cases?

follow up PR is to drop the bootargs stuff from kairos directly and just use the things coming form the package, i.e. not have any special thing done on our repo and make the framework provide everything needed. This should be the last step needed to clean the kairos repo from special things :)

PR #477, opened 2023-10-10
Make kairos-sysext be generic enough to include on all flavors

Itxaka:
It will not install the systemd-sysext package if the binary is not there

### DIFF
--- a/packages/dracut/sysext/files/90kairos-sysext/module-setup.sh
+++ b/packages/dracut/sysext/files/90kairos-sysext/module-setup.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+# This module tries to add systemd-sysext module to the initramfs if its in the system, otherwise it does nothing
+
+# called by dracut
+check() {
+    return 0
+}
+
+# called by dracut
+depends() {
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries systemd-sysext || return 1
+    # Check if the module files exists
+    # This is not normal but on ubuntu-22 the binary for sysext exists but the dracut module doesnt, so we
+    # need to do further checks before including it otherwise it will fail to build the initrd
+    files=( "${dracutbasedir}"/modules.d/??systemd-sysext )
+    [ "${#files[@]}" -ge 2 ] && return 1
+    if [ -d "${files[0]}" ]; then
+        echo "systemd-sysext"
+        return 0
+    else
+        dinfo "module 'systemd-sysext' will not be installed, because module dir could not be found!"
+    fi
+    return 1
+}
+
+# called by dracut
+installkernel() {
+    return 0
+}
+
+# called by dracut
+install() {
+    return 0
+}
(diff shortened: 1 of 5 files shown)


Itxaka:
Test ongonig for this: https://github.com/kairos-io/kairos/pull/1904

Itxaka:
Expected: On systems that do not support systemd-sysext becuase they are older, its not installed, on systems that support it, it is.

Outcome of the test:

on ubuntu-20-lts (no sysext available): `+base-image | dracut: dracut module 'kairos-sysext' will not be installed, because command 'systemd-sysext' could not be found!`
on ubuntu-22-lts (no sysext available): `systemd-sysext` module was NOT added to initramfs. Silent skip due to the comments in the code, mainly it has the binary but not the dracut module, so we add extra checks and fail in a different place.

on opensuse-tumbleweed (sysext available),  `systemd-sysext` module was added to initramfs
on opensuse-leap (sysext available),  `systemd-sysext` module was added to initramfs
on ubuntu (sysext available),  `systemd-sysext` module was added to initramfs
on almalinux (sysext available),  `systemd-sysext` module was added to initramfs
on fedora (sysext available),  `systemd-sysext` module was added to initramfs
on debian (sysext available),  `systemd-sysext` module was added to initramfs
on rockylinux (sysext available),  `systemd-sysext` module was added to initramfs



So as a side effect of this generic approach of adding if its there, we GAIN sysext support in all newer distros :D

packages/dracut/sysext/files/90kairos-sysext/module-setup.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+# This module tries to add systemd-sysext module to the initramfs if its in the system, otherwise it does nothing
+
+# called by dracut
+check() {
+    return 0
```
Itxaka, line 6: This indicates to dracut to always include the module

packages/dracut/sysext/files/90kairos-sysext/module-setup.sh
```diff
@@ -0,0 +1,35 @@
...
+depends() {
+    # If the binary(s) requirements are not fulfilled the module can't be installed.
+    require_binaries systemd-sysext || return 1
+    # Check if the module files exists
+    # This is not normal but on ubuntu-22 the binary for sysext exists but the dracut module doesnt, so we
+    # need to do further checks before including it otherwise it will fail to build the initrd
+    files=( "${dracutbasedir}"/modules.d/??systemd-sysext )
+    [ "${#files[@]}" -ge 2 ] && return 1
+    if [ -d "${files[0]}" ]; then
+        echo "systemd-sysext"
```
Itxaka, line 19: this is what brings the systemd-sysext module as dependent, so it includes on systems that have botht he binary and the dracut module available, otherwise this module depends on nothing and its basically an empty thing that does nothing.

jimmykarily (approved):

mauromorales (approved):
very nice

PR #478, opened 2023-10-10
Add missing cryptsetup binary to alpine initrd

Itxaka:
Not bumping becuase no kernel change, will come later


Itxaka:
Merging directly, this does not bump any packages, its a initrd change for alpine pure

PR #483, opened 2023-10-10
Add dm_mod to initramfs for mapper devices

Itxaka:


Itxaka:
only for alpine initrd, merging directly cc @kairos-io/maintainers

PR #492, opened 2023-10-16
Even more alpine fixes

Itxaka:
- drop cgroups config, defaults are better
 - take over ttys on livecd, installer and reset


mauromorales (approved):
nice

PR #495, opened 2023-10-17
Enable network by default on alpine initramfs

Itxaka:
Can be disabled with rd.neednet=0
Make so no failure if we get no network, we tried our best Make udhcpc not block boot


mauromorales (commented):
lgtm

PR #496, opened 2023-10-17
Dont break inittab completely

Itxaka:
Append to it instead of creating it from scratch as it has some defaults that are needed for the correct working of the system like openrc starting and so on


Itxaka:
merging directly as its needed for alpine

PR #503, opened 2023-10-19
Move kernel+initrd package to provide only scripts

Itxaka:
So we can build it as part of the build process


mauromorales (approved):
lgtm

PR #505, opened 2023-10-19
Add and enable openrc/systemd service via yip files

Itxaka:
Instead of having them hardcoded as a package


packages/static/kairos-overlay-files/files/etc/systemd/system/systemd-firstboot.service
```diff
@@ -0,0 +1 @@
+/dev/null
```
mauromorales, line 1: what is this about?

Itxaka: Masking the service by linking it to /dev/null which is what systems does when you mask a service.

We don't want the first boot service to start and block boot boys asking stupid questions so we need to mask it.

jimmykarily (approved):

PR #522, opened 2023-10-25
Remove the +k3s1 part on k3s packages

mauromorales:
:warning: the latest extensions for luet need to be released for this to work https://github.com/Luet-lab/luet-repo/commit/b6b4c53511b214aa2032c1cf14644a522759ce28

relates to kairos-io/kairos#1934


Itxaka (approved):

mauromorales:
ok after talking about this in the daily I realised there's a problem. If I make them "release" strategy, then they will all get the latest release. But if I leave them without strategy, it will show rc releases. So I might need to change it back to without strategy but filter out rcs.

PR #535, opened 2023-11-08
Introduce KAIROS_ARTIFACT in bootargs.cfg and os-release

jimmykarily:
so store the full bootable artifact name as created by the naming.sh script in the kairos repo.

This needs a change in the kairos Earthfile (OSRELEASE target) to add a value to this variable


packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
```diff
@@ -15,11 +15,11 @@ function setExtraConsole {
     source (loop0)/etc/os-release
     set baseExtraConsole="console=ttyS0"
     # rpi
-    if regexp "arm-rpi" $KAIROS_FLAVOR; then
+    if regexp "arm64-rpi" $KAIROS_ARTIFACT; then
```
Itxaka, line 18: I think we could drop the arm stuff in here and orin, no way its gonna be x86 no?

PR #544, opened 2023-11-21
Make tty1 the default console and let kairos service logs there

jimmykarily:
The order in the cmdline matters and it seems that the last one defined is the "default" console. Setting StandardOutput to journal+console makes `journalctl -u kairos` print the installer logs (maintaining the existing behaviour of printing them in tty1)

This should allow us to refactor the test here: https://github.com/kairos-io/kairos-agent/pull/174#issuecomment-1819318236

because then we will have access to the installer logs using `journalctl`

### DIFF
--- a/packages/static/kairos-overlay-files/collection.yaml
+++ b/packages/static/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.1.12"
+    version: "1.1.13"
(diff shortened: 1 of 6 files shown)


Itxaka:
As long as it doesn't break anything, pretty good catch!

jimmykarily:
> As long as it doesn't break anything, pretty good catch!

Well, let's merge, bump in kairos and see. I don't see any different in the previous behavior with my manual test so I assume it's fine. It should only add the journalctl functionality without removing anything. Let's see.

PR #553, opened 2023-12-01
Restore grub artifacts to leap

Itxaka:
tumbleweed seems to have some issues


mauromorales (approved):
change lgtm, but wasn't there an issue in leap about arm and amd ... that it used to ship all of them on the amd package or something like that?

Itxaka:
> change lgtm, but wasn't there an issue in leap about arm and amd ... that it used to ship all of them on the amd package or something like that?

mostly that tumbleweed didnt ship both arches while leap did. But doesnt matter becuase now the build is done with templates depending on the arch

PR #557, opened 2023-12-04
Fix services tty output

Itxaka:
- interactive installer: output was not connected to tty so it was impossible to answer any questions
 - reset: was not getting the output and taking over the tty so it got overwritten by other output. It could be stopped in the middle
 - installer: same as above but it could lead to stopping the install in the middle of it
 - recovery: same as above, QR code didnt even display

### DIFF
--- a/packages/static/kairos-overlay-files/files/etc/systemd/system/kairos-interactive.service
+++ b/packages/static/kairos-overlay-files/files/etc/systemd/system/kairos-interactive.service
@@ -3,7 +3,7 @@ Description=kairos interactive-installer
 After=sysinit.target
 [Service]
 Type=oneshot
-# imput/output to tty as its interactive
+# input/output to tty as its interactive
 # otherwise it will be silent and with no input
 StandardInput=tty
 StandardOutput=tty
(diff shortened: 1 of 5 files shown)


packages/static/kairos-overlay-files/files/etc/systemd/system/kairos-interactive.service
```diff
@@ -3,12 +3,15 @@ Description=kairos interactive-installer
 After=sysinit.target
 [Service]
 Type=oneshot
+# imput/output to tty as its interactive
```
mauromorales, line 6:
```suggestion
# input/output to tty as its interactive
```

mauromorales (approved):
lgtm

Itxaka:
I don't get why it doesn't work, probably systemd is mangling the output somehow but it doesn't. You can try yourself by testing master with interactive install or recovery and see how it breaks the tty1 output. With this, which was the original setting from a long time ago, it doesn't.

But to make it work and not keep it broken we need to reverse this to a working state and then change it back to jornal+console as currently is a regression.

I wonder if the way we interact by Kairos agent is the reason, maybe we need to do changes there.

jimmykarily:
I'm fine to revert it for now to fix the regression. Let's open a new issue to investigate further and bring the change back properly.

PR #559, opened 2023-12-05
Fix services output

Itxaka:
- make sure getty is stopped and masked in case other service tries to bring it up when we are running reset,recovery or installers
 - make systemd nto show the output in the main tty by default so it doesnt pollute the tty with messages afterwards

### DIFF
--- a/packages/static/kairos-overlay-files/collection.yaml
+++ b/packages/static/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.1.14"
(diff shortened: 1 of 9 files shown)


Itxaka:
We could also send a signal to PID1 directly in the Pre exec to stop logging messages if we really want those messages to appear during boot?

https://www.freedesktop.org/software/systemd/man/latest/systemd.html#SIGRTMIN+21

Itxaka:
And actually that seems to work much nicer, it shows the start messages from systemd but stops them as soon as we start our services.... changing it to that!

jimmykarily (approved):

PR #587, opened 2023-12-20
Enable MAC as DHCP client identifier

kreeuwijk:
This PR adjusts the default DHCP configuration to use the Client Identifier for DHCP from a random GUID to the NIC's MAC address. This brings it in line with every other OS and customer expectations. It ensures that the Edge device receives the same DHCP lease from the DHCP server as long as the lease is still valid.

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/05_network.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/05_network.yaml
@@ -18,6 +18,8 @@ stages:
             Name=en*
             [Network]
             DHCP=yes
+            [DHCP]
+            ClientIdentifier=mac
         - path: /etc/systemd/network/20-dhcp-legacy.network
           permissions: 0644
           owner: 0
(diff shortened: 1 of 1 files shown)


Itxaka (approved):

Itxaka:
In order to make this work you would need to bump the package version as well, otherwise it wont be build :)

kreeuwijk:
@Itxaka I wouldn't know how to do that?

Itxaka:
It's here, on the package directory, the version number need to be bumped so it gets build and pushed to the repos, otherwise it won't be build as a package with the same version is already published https://github.com/kairos-io/packages/blob/main/packages%2Fstatic%2Fkairos-overlay-files%2Fcollection.yaml

Itxaka:
Ill bump the version myself after merging this, thanks @kreeuwijk !

kreeuwijk:
Ah I see, thanks!

PR #590, opened 2023-12-28
Fix version check on firmwares

Itxaka:
upstream changed the compression format if the repodata so it was failing


Itxaka:
cc @mauromorales merging directly!

PR #597, opened 2023-12-29
Move grub provider to tumbleweed

Itxaka:
This bumps grub to 2.12 which should bring fixes

Potentially this can bring the tpm out of memory issue back, but I would like to retest it as it brings a lot of fixex and improvements to grub, potentially including secureboot fixes

### DIFF
--- a/packages/livecd/grub2/build.yaml
+++ b/packages/livecd/grub2/build.yaml
@@ -1,4 +1,4 @@
-image: "opensuse/leap:15.5"
+image: {{.Values.image}}
 
 prelude:
 {{if .Values.arch }}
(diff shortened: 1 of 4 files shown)


mauromorales (approved):

mudler:
changing grub version is scary - not sure if worth going into https://github.com/kairos-io/kairos/issues/1842 again, or it got fixed upstream?

Itxaka:
> changing grub version is scary - not sure if worth going into [kairos-io/kairos#1842](https://github.com/kairos-io/kairos/issues/1842) again, or it got fixed upstream?

mainly for testing as for livecd there is still not a solution as the i386 package is still required to build hybrid isos :(

Itxaka:
Im stupid. We are using the shim+efi of leap 15.5 so it has the proper signatures and they boot, but once you load the kernel, unless its leap 15.5 kernel, its signed by a different entity (like red hat or ubuntu) so signatures dont match.

This is why it fails.....

PR #604, opened 2024-01-04
Increase fs.inotify.max_user_instances sysctl value

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/2071


mauromorales (approved):
I'd say let's go with this, seems like a good default and we can adapt if anyone opens an issue

PR #607, opened 2024-01-05
Disable tpm module on efi for grub

Itxaka:
It runs out of memory when loading the loop device as it tries to measure the full file by loading it into memory

Part of supposrting secureboot in all flavors

### DIFF
--- a/packages/static/grub-config/files/grub.cfg
+++ b/packages/static/grub-config/files/grub.cfg
@@ -59,6 +59,10 @@ insmod squash4
 insmod serial
 insmod regexp
 loadfont unicode
+if [ "${grub_platform}" = "efi" ]; then
+    ## workaround for grub2-efi bug: https://bugs.launchpad.net/ubuntu/+source/grub2/+bug/1851311
+    rmmod tpm
+fi
 
 menuentry "${display_name}" --id cos {
   search --no-floppy --label --set=root COS_STATE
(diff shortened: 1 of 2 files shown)


packages/static/grub-config/files/grub.cfg
```diff
@@ -59,6 +59,10 @@ insmod squash4
 insmod serial
 insmod regexp
 loadfont unicode
+if [ "${grub_platform}" = "efi" ]; then
+    ## workaround for grub2-efi bug: https://bugs.launchpad.net/ubuntu/+source/grub2/+bug/1851311
+    rmmod tpm
```
mauromorales, line 64: does that mean that we cannot use tpm encryption with EFI atm?

mauromorales (approved):

packages/static/grub-config/files/grub.cfg
Itxaka: This is only for measurement of the artifacts into the tpm registers. Because for now we only support measured boot under uki which uses systemd-boot this should not affect us. 

Let's hope that this is fixed if we ever want to use those measurements for something with grub 🤣

Btw this only affects debian derivatives. I hope this is fixed once debian bumps to the 2.12 version

mauromorales: gotcha, thanks for clarifying

PR #608, opened 2024-01-09
uki fixes for cc files

Itxaka:


packages/static/kairos-overlay-files/files/system/oem/08_grub.yaml
```diff
@@ -100,11 +102,12 @@ stages:
             fi
     - &grubinstall
       name: "Grub branding"
-      if: '[ -e "/etc/kairos/branding/grubmenu.cfg" ]'
+      if: '[ -e "/etc/kairos/branding/grubmenu.cfg" ] && grep -vq "rd.immucore.uki" /proc/cmdline'
```
mauromorales, line 105: i wonder if there's a way to group these, like having small scripts so we don't forget to change in one place

mauromorales (approved):

packages/static/kairos-overlay-files/files/system/oem/08_grub.yaml
Itxaka: YES. small functions would be awesome

I guesswe could use the yaml anchors thing to not repeat ourselves somehow?

PR #617, opened 2024-01-12
Drop linuxefi/initrdefi

Itxaka:
Is not needed anymore, grub is smart enough to use whats needed


mudler (approved):
can confirm, usage of linuxefi initrdefi is quite old, not sure how it got in here

PR #619, opened 2024-01-12
Drop regex module from bootargs

Itxaka:


packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
```diff
@@ -15,11 +16,11 @@ function setExtraConsole {
     source (loop0)/etc/os-release
     set baseExtraConsole="console=ttyS0"
     # rpi
-    if regexp "arm64-rpi" $KAIROS_ARTIFACT; then
+    if test $KAIROS_MODEL == "rpi3" -o test $KAIROS_MODEL == "rpi4"; then
```
Itxaka, line 19: now that we have the model on its own var, this is much easier than before as we can exact match!

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
```diff
@@ -1,7 +1,8 @@
 function setSelinux {
     source (loop0)/etc/os-release
     set baseSelinuxCmd=""
-    if regexp "rockylinux|fedora|almalinux|redhat|tumbleweed" $KAIROS_ARTIFACT; then
+    # opensuse does not bundle regexp module so if we are on leap or tumbleweed we need to do this manually
+    if test $KAIROS_FLAVOR == "rockylinux" -o test $KAIROS_FLAVOR == "fedora" -o test $KAIROS_FLAVOR == "almalinux" -o test $KAIROS_FLAVOR == "redhat" -o test $ID == "opensuse-tumbleweed"; then
```
jimmykarily, line 5: We can stick to `KAIROS_` prefixed variables, which we fully  control, if you use `$KAIROS_FLAVOR_RELEASE == "tumbleweed"` instead of `ID`.

jimmykarily (approved):
Nice!

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
```diff
@@ -1,7 +1,8 @@
 function setSelinux {
     source (loop0)/etc/os-release
     set baseSelinuxCmd=""
-    if regexp "rockylinux|fedora|almalinux|redhat|tumbleweed" $KAIROS_ARTIFACT; then
+    # opensuse does not bundle regexp module so if we are on leap or tumbleweed we need to do this manually
```
jimmykarily, line 4: this comment will be confusing after we merge this PR. Anyone viewing this file won't understand why we explain the non-usage of `regexp`.

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
mauromorales, line 5: Can we use FAMILY instead?

```suggestion
    if test $KAIROS_FAMILY == "rhel" -o test $ID == "opensuse-tumbleweed"; then
```

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
mauromorales: very cool

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
```diff
@@ -1,7 +1,7 @@
 function setSelinux {
     source (loop0)/etc/os-release
     set baseSelinuxCmd=""
-    if regexp "rockylinux|fedora|almalinux|redhat|tumbleweed" $KAIROS_ARTIFACT; then
+    if test $KAIROS_FAMILY == "rhel" -o test $ID == "opensuse-tumbleweed"; then
```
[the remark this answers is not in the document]
mauromorales: this one?

> this comment will be confusing after we merge this PR. Anyone viewing this file won't understand why we explain the non-usage of regexp.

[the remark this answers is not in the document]
mauromorales: ah no, this one I guess:

> We can stick to KAIROS_ prefixed variables, which we fully control, if you use $KAIROS_FLAVOR_RELEASE == "tumbleweed" instead of ID.

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
Itxaka: but do we have that variable already or you think we should introduce it? I thin what its there should be enough to identify, especially this as this is tumbleeed only

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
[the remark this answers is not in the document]
jimmykarily: sorry, yes. These are both comments I made :D .

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
jimmykarily: It's there: https://github.com/kairos-io/kairos-sdk/blob/c21aa7f149cbbea5416f7fd475d9c6613d208d4e/versioneer/versioneer.go#L18 . But it's ok as it is too. It's minor.

PR #636, opened 2024-01-26
Run installer on uki livecd mode and use the new init detection

Itxaka:


packages/static/kairos-overlay-files/files/system/oem/02_agent.yaml
```diff
@@ -1,11 +1,11 @@
 name: "Start agent"
 stages:
   boot:
-    - if: '[ ! -f "/run/cos/recovery_mode" ] && [ -d "/usr/share/systemd" ]'
+    - if: '[ ! -f "/run/cos/recovery_mode" ] && [ "$(kairos-agent state get kairos.init)" == "systemd" ]'
```
mauromorales, line 4: not for now, but seeing this, makes me feel like we could simply have some querying mechanism in the agent e.g.

```
kairos-agent state is kairos.init.systemd
echo $?
1
```

packages/static/kairos-overlay-files/files/system/oem/52_installer.yaml
```diff
@@ -31,11 +43,14 @@ stages:
         - echo "tty1::respawn:/usr/bin/kairos-agent interactive-install --shell tty1" >> /etc/inittab
   boot:
     - if: |
-          [ "$(kairos-agent state get boot)" == "livecd_boot" ] && [ -f "/sbin/rc-service" ]
+        [ "$(kairos-agent state get boot)" == "livecd_boot" ] && [ -f "/sbin/rc-service" ]
```
mauromorales, line 46: should there not be a 

```
"$(kairos-agent state get kairos.init)" == "openrc"
```

here instead?

Itxaka: mmhm yeah, good point. Becuase openrc is "simpler" to check I forgot about using the state xD

Itxaka: done!

packages/static/kairos-overlay-files/files/system/oem/02_agent.yaml
Itxaka: umm, not sure. State is state right? do we want to introduce that kind of operations onto it? would be cool indeed and we could piggyback the existing state I guess, but seems like a can of worms hahaha

mauromorales (approved):
LGTM

PR #639, opened 2024-01-30
Change nodepair.enable by install-mode

Itxaka:


packages/static/kairos-overlay-files/files/system/oem/52_installer.yaml
```diff
@@ -1,43 +1,31 @@
...
+    - if: grep -q "install-mode" /proc/cmdline && grep -vq "install-mode-interactive" && ([ "$(kairos-agent state get boot)" == "livecd_boot" ] || [ -f /run/cos/uki_install_mode ]) && [ "$(kairos-agent state get kairos.init)" == "systemd" ]
       commands:
         - systemctl disable getty@tty1
         - systemctl stop getty@tty1
         - systemctl mask getty@tty1
         - systemctl enable kairos
         - systemctl enable kairos-webui
     # Starts installer on boot for openRC based systems
-    - if: grep -q "nodepair.enable" /proc/cmdline && [ "$(kairos-agent state get kairos.init)" == "openrc" ]
+    - if: grep -q "install-mode" /proc/cmdline && grep -vq "install-mode-interactive" && [ "$(kairos-agent state get kairos.init)" == "openrc" ]
```
mauromorales, line 15: I think this produces a false positive `-v` will be successful whether "install-mode-interactive" is present or not

mauromorales (changes requested):

packages/static/kairos-overlay-files/collection.yaml
```diff
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.1.22"
+    version: "1.1.24"
```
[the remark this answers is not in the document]
Itxaka: umm

this was supposed to be bumped in a previous PR...

packages/static/kairos-overlay-files/files/system/oem/52_installer.yaml
Itxaka: why? Seems to work here:

```
# itxaka @ zeus in ~/projects/packages on git:nodepair_enable x [15:42:41] C:1
$ cat install.txt  
install-mode

# itxaka @ zeus in ~/projects/packages on git:nodepair_enable x [15:42:44] 
$ cat install-interactive.txt 
install-mode install-mode-interactive

# itxaka @ zeus in ~/projects/packages on git:nodepair_enable x [15:42:47] 
$ (grep -q install-mode install.txt && grep -vq install-mode-interactive install.txt)           

# itxaka @ zeus in ~/projects/packages on git:nodepair_enable x [15:42:57] 
$ echo $?
0

# itxaka @ zeus in ~/projects/packages on git:nodepair_enable x [15:42:58] 
$ (grep -q install-mode install-interactive.txt && grep -vq install-mode-interactive install-interactive.txt)

# itxaka @ zeus in ~/projects/packages on git:nodepair_enable x [15:43:06] C:1
$ echo $?
1
```

install.txt represents the cmdline of the defaul;t install mode, the cmdline that is on kairos by defaultt
install.txt represents the cmdline of uki with interactive install, which has both install-mode and install-mode-interactive

We want to only trigger the kairos install service if install-mode is in the cmdline but NOT install-mode-interactive, as for that the interactive installer service should trigger instead.


Where are you seeing the issue @mauromorales ? I cant find it and my local tests seems to confirm this (hint: is how I got this line, by testing it locally to make sure it worked lmao)

packages/static/kairos-overlay-files/collection.yaml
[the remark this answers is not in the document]
Itxaka: now its updated, sorry :D

packages/static/kairos-overlay-files/files/system/oem/52_installer.yaml
mauromorales: sorry, my bad, tested again and working as expected. I got confused

mauromorales (approved):
this is starting to look like LISP ((((())))) 

i wonder if we could just somehow have shared scripts that end up looking like `if: install-mode-only.sh` or so or if it's a bad idea

Itxaka:
> this is starting to look like LISP ((((()))))
> 
> i wonder if we could just somehow have shared scripts that end up looking like `if: install-mode-only.sh` or so or if it's a bad idea

we did discuss the possibility of adding templates and or minicommands no? to yip directly. Maybe we could have some helpers.tpl or something like that that contain functions to use. But at that point we are starting to create a mini language on itself lmao.

Maybe we could leverage the helm templates stuff, they do have functions and such lol

PR #667, opened 2024-02-07
Add header to nvidia generated configs

mauromorales:
this on its own does not fix the issue, but it will be needed

relates to kairos-io/kairos#2221

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/12_nvidia.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/12_nvidia.yaml
@@ -11,9 +11,9 @@ stages:
         permissions: 0644
         owner: 0
         group: 0
-      # TODO: This is no longer read. This should be moved to cloud-init file.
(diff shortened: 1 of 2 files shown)


packages/static/kairos-overlay-files/files/system/oem/12_nvidia.yaml
```diff
@@ -11,9 +11,9 @@ stages:
         permissions: 0644
         owner: 0
         group: 0
-      # TODO: This is no longer read. This should be moved to cloud-init file.
```
jimmykarily, line 14: This is still true, no?

jimmykarily (approved):

packages/static/kairos-overlay-files/files/system/oem/12_nvidia.yaml
mauromorales: for the time being but it's what we need to fix in this story

PR #668, opened 2024-02-08
Make sure that if a package file has been changed, that a related definition/collection file has also been changed

mauromorales:
Every time I touch a package I forget to bump the version, so I thought this would be helpful


tools/bump-validator/main.go
```diff
@@ -0,0 +1,50 @@
...
+	return basename == "definition.yaml" || basename == "collection.yaml"
+}
+
+// reduceRelated keeps only the definition files that are siblings or in a parten directory of the given file
+func reduceRelated(file string, files []string) []string {
+	var relatedFiles []string
+	dir := filepath.Dir(file)
+	for _, f := range files {
+		otherDir := filepath.Dir(f)
+		if isDefinitionFile(f) && (strings.Contains(dir, otherDir) || strings.Contains(otherDir, dir)) {
```
jimmykarily, line 22: Maybe use `strings.HasPrefix` to avoid matching `otherDir` in random location in `dir` (edge case)

tools/bump-validator/main.go
```diff
@@ -0,0 +1,50 @@
...
+		// fmt.Println("Processing file ", file)
+		if isDefinitionFile(file) {
+			// fmt.Println("Skipping definition file ", file)
+			continue
+		}
+
+		relatedFiles := reduceRelated(file, files)
+		if len(relatedFiles) == 0 {
+			missingVersionBump = true
+			fmt.Println("Error: Version bump missing for file ", file)
```
jimmykarily, line 43: and `break` ?

jimmykarily: or even `os.Exit(1)` here? I don't see how `missingVersionBump` could turn to false again. What's the reason to continue the for loop?

jimmykarily: oh the `Println` is the one we want to happen. Gotcha, thanks.

PR #689, opened 2024-02-16
Provide a package with systemd-ukify+systemd-measure

Itxaka:
Should be good enough to install in the osbuilder in order to build uki files from different base OS

### DIFF
--- a/packages/system/systemd/build.yaml
+++ b/packages/system/systemd/build.yaml
@@ -3,7 +3,7 @@ image: {{ .Values.image }}
 package_dir: /package
 
 prelude:
-  - dnf update -y && dnf install -y gcc meson git ninja-build gperf libpcap libcap libcap-devel cmake libmount-devel python3-jinja2 rsync diffutils python3-pyelftools openssl-devel
+  - dnf update -y && dnf install -y gcc meson git ninja-build gperf libpcap libcap libcap-devel cmake libmount-devel python3-jinja2 rsync diffutils python3-pyelftools openssl-devel tpm2-*
   - mkdir -p src/
   - PACKAGE_VERSION=${PACKAGE_VERSION%\+*} && cd src/ && git clone --branch v${PACKAGE_VERSION} https://github.com/systemd/systemd.git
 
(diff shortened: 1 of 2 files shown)


packages/system/systemd/build.yaml
```diff
@@ -9,19 +9,32 @@ prelude:
...
...=false -Dinitrd=false -Dimportd=false -Dkernel-install=false -Dlocaled=false -Dmachined=false -Dnetworkd=false -Dnss-myhostname=false -Dnss-mymachines=false -Dnss-systemd=false -Doomd=false -Dportabled=false -Dhwdb=false -Dpstore=false -Dquotacheck=false -Drandomseed=false -Drepart=false -Dresolve=false -Drfkill=false -Dsysext=false -Danalyze=false -Dsysupdate=false -Dsysusers=false -Dstoragetm=false -Dtimedated=false -Dtimesyncd=false -Dtmpfiles=false -Duserdb=false -Dvconsole=false -Dxdg-autostart=false -Didn=false -Dpolkit=false -Dnscd=false -Dkmod=false -Ddbus=false -Dglib=false -Dbacklight=false -Dldconfig=false -Dgshadow=false -Dwheel-group=false -Dadm-group=false -Dxkbcommon=false -Dzstd=false -Dlz4=false -Dutmp=false -Dlink-udev-shared=false -Dlink-systemctl-shared=false -Dlink-networkd-shared=false -Dlink-timesyncd-shared=false -Dlink-journalctl-shared=false -Dlink-boot-shared=false -Dlink-portabled-shared=false -Denvironment-d=false -Dqrencode=false -Dpwquality=false -Dlibcurl=false -Dfdisk=false -Dlibidn2=false -Dlibiptc=false -Ddns-over-tls=false -Didn=false -Dgnutls=false -Dp11kit=false -Dlibidn=false -Dlibidn2=false -Dgcrypt=false -Dxz=false -Dzlib=false -Dbzip2=false
```
mauromorales, line 14: no clue what changed here :/

mauromorales (approved):

packages/system/systemd/build.yaml
Itxaka: its on the comment above, to make meson build the systemd-measure binary we need to enable:
 - openssl (-Dopenssl=true)
 - blkid (-Dblkid=true)
 - bootloader, was already enable to build the efi artifacts (-Dbootloader=true)

Then to generate a custom shared lib with a kairos name on it so it doesnt collide with installed libs we set `-Dshared-lib-tag=${PACKAGE_VERSION}-kairos` which will generate a `libsystemd-shared-255-kairos.so` which the systemd-measure binary will look for

I moved all those options to be at the start of the command so they are "easy" to check xD

mauromorales: gotcha, thanks for clarifying

PR #706, opened 2024-02-21
2275 kubernetes no pivot

jimmykarily:
Part of https://github.com/kairos-io/kairos/issues/2275

### DIFF
--- a/packages/k8s/k3s/build.yaml
+++ b/packages/k8s/k3s/build.yaml
@@ -32,12 +32,16 @@ steps:
   - rm -rf installer.sh
   - chmod +x /usr/bin/k3s
   - TAG=$INSTALL_K3S_VERSION ./fetch-template.sh
-  - mv config.toml.tmpl /etc/k3s-config.toml.tmpl
+  - mv config.toml.tmpl /usr/share/config/kairos/k3s/containerd-config.toml.tmpl
 
 includes:
 - ^/usr/bin/k3s
-- ^/etc$
-- ^/etc/k3s-config.toml.tmpl$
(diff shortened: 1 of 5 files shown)


packages/k8s/k3s/build.yaml
```diff
@@ -31,9 +31,11 @@ steps:
   - bash installer.sh agent
   - rm -rf installer.sh
   - chmod +x /usr/bin/k3s
+  - TAG=$INSTALL_K3S_VERSION ./fetch-template.sh
 
 includes:
 - ^/usr/bin/k3s
+- ^config.toml.tmpl
```
mudler, line 38: as a note: this would be installed over /config.toml.tmpl right? maybe we should default to some other dir, e.g. /usr/share/config/kairos/k3s/...

jimmykarily: yes, good point. I was just playing around with the script. Moved it now.

packages/static/kairos-overlay-files/files/system/oem/53_k3s.yaml
```diff
@@ -0,0 +1,6 @@
+name: "k3s configuration"
+stages:
+  boot:
+    - if: '[ ! -f "/run/cos/recovery_mode" ] && [ ! -e "/run/cos/uki_install_mode" ] && [[ $(source /etc/os-release; echo "$KAIROS_VARIANT") == "standard" ]]'
```
jimmykarily, line 4: I'm not sure if mixing double `[[` with `[` conditions works

mudler (approved):

PR #709, opened 2024-02-22
Ship ext2 drivers with the systemd-boot package

Itxaka:


packages/system/systemd/build.yaml
```diff
@@ -13,10 +13,12 @@ steps:
   - cp /usr/lib/systemd/boot/efi/systemd-bootaa64.efi /package/usr/kairos/
   - cp /usr/lib/systemd/boot/efi/linuxaa64.efi.stub /package/usr/kairos/
   - cp /usr/lib/systemd/boot/efi/addonaa64.efi.stub /package/usr/kairos/
+  - curl -L https://github.com/pbatard/efifs/releases/download/v1.9/ext2_aa64.efi -o /package/usr/kairos/ext2_aa64.efi
```
mudler, line 16: Can't we compile them? 

https://github.com/pbatard/efifs?tab=readme-ov-file#bonus-commands-to-compile-efifs-using-edk2-on-a-vanilla-debian-gnulinux-10x

I think at this stage pulling things randomly from github is pushing it a little bit

Itxaka: well yeah we can I guess?

Itxaka: thats gonna take hours btw

jimmykarily: since the version is hardcoded, maybe check a sha1 signature at least? As in `this is a verified artifacts that we have checked`.

Itxaka:
I am unable to do the aarch64 compilation properly, so leaving it like this?

jimmykarily:
> I am unable to do the aarch64 compilation properly, so leaving it like this?

I think that's ok for this first iteration. Unless the product needs aarch64 for some reason.

mudler:
aarch64 would be toothless now, no TPM, no fun

PR #710, opened 2024-02-26
Reduce binary sizes

mauromorales:
relates to: kairos-io/kairos#2285

### DIFF
--- a/packages/toolchain-go/build.yaml
+++ b/packages/toolchain-go/build.yaml
@@ -1,11 +1,19 @@
 {{ if .Values.variant }}
-image: "golang:{{.Values.version}}-{{.Values.variant}}"
+image: "golang:{{.Values.tag}}-{{.Values.variant}}"
 {{ else }}
-image: "golang:{{.Values.version}}"
+image: "golang:{{.Values.tag}}"
 {{ end }}
 
-{{ if eq .Values.variant "alpine" }}
 prelude:
+{{ if eq .Values.variant "alpine" }}
 - apk update
 - apk add curl make git bash
+{{ end }}
+{{ if eq .Values.variant "bookworm" }}
+- apt-get update && apt-get install -y curl xz-utils
+- curl -LO https://github.com/upx/upx/releases/download/v4.2.2/upx-4.2.2-amd64_linux.tar.xz
+- tar -xvf upx-4.2.2-amd64_linux.tar.xz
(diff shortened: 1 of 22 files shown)


Itxaka (approved):
No immucore, k9s, k3s etc all?

Itxaka:
Btw I think kceypt is not being shipped as part of Kairos anymore as everything is done via it's public apis directly on immucore and agent

mauromorales:
@Itxaka good point, will add those

Itxaka:
> @Itxaka good point, will add those

Im a bit worried about immucore as it runs on an empty system no idea what upx needs to be able to autodecompress?

packages/toolchain-go/build.yaml
```diff
@@ -1,11 +1,19 @@
...
+image: "golang:{{.Values.tag}}"
 {{ end }}
 
-{{ if eq .Values.variant "alpine" }}
 prelude:
+{{ if eq .Values.variant "alpine" }}
 - apk update
 - apk add curl make git bash
+{{ end }}
+{{ if eq .Values.variant "bookworm" }}
```
mudler, line 12: :+1: 

unbelievelable, only bookworm doesn't have upx. It is such an historic package - I'm a bit shocked :)

https://packages.debian.org/search?searchon=names&keywords=upx

mudler (approved):

packages/toolchain-go/build.yaml
mauromorales: yup I was a bit confused there

packages/toolchain-go/build.yaml
```diff
@@ -1,11 +1,19 @@
...
+- apt-get update && apt-get install -y curl xz-utils
+- curl -LO https://github.com/upx/upx/releases/download/v4.2.2/upx-4.2.2-amd64_linux.tar.xz
```
mudler, line 14: of course this doesn't work for arm64, need to handle the architecture properly here

mudler:
aren't we missing here to compress/minimize immucore too?

mauromorales:
@Itxaka ok I'll do immucore last and in a different PR

mauromorales:
@mudler yes my bad, will add a bunch of others I completely forgot we had there. Also, will remove the docs since they account for half of the size of the compressed binary (12Mb and 24 total)

PR #735, opened 2024-03-06
Revert toml template for k3s

Itxaka:
Issue should be fixed in undelying OS now


packages/k8s/k3s/collection.yaml
```diff
@@ -1,7 +1,7 @@
 packages:
   - name: k3s-openrc
     category: k8s
-    version: "1.29.0+2"
```
mauromorales, line 4: why not +3? in the end it doesn't really matter but I think ok to keep different build versions for openrc and systemd and I did some changes to the pipeline so it can handle them

Itxaka: OCD basically, I could not stand them being in a different revision

mauromorales (approved):

PR #755, opened 2024-03-18
Fix auto reset on alpine

Itxaka:
Agent needs to know that we are running unnatended AND we want to reboot when its done for the auto reset to behave as expected

Fixes: https://github.com/kairos-io/kairos/issues/2136


Itxaka:
Confirmed that it fixes the issue

jimmykarily (approved):

PR #761, opened 2024-03-20
Move the fips toolchain-go-ubuntu back to ubuntu

Itxaka:


mauromorales (approved):

mauromorales:
I think this might break upx though, but maybe we can just not run it for fips?

Itxaka:
> I think this might break upx though, but maybe we can just not run it for fips?

looks like immucore is not being upx-ed somehow? So it should be ok for it. But for other yeah, we migth need either an alternatice or build a upx package or something like that

Itxaka:
mmh, upx is on bullseye? so it should be no issue other than we need to also install it in here

Itxaka:
most importatn thing here is to check that the fips/immucore package when build, bring the exact `>=1.19.10+1` version which should fix the issue, otherwise if it bring a lower version that means its not being built with the required toolchain

Itxaka:
looks like it did correctly:

```
 INFO   📦  Compiling fips/immucore-0.1.21+2 .... ☕ 
 INFO   🌳  Build dependencies for fips/immucore-0.1.21+2
 INFO    ↪  fips/toolchain-go-ubuntu-1.19.10+1 🍃 
 INFO   📦  1/1 fips/immucore-0.1.21+2 ⤑ 🔨  build fips/toolchain-go-ubuntu-1.19.10+1  starts
 INFO   SUCCESS 📦  1/1 fips/immucore-0.1.21+2 ⤑ 🔨  build fips/toolchain-go-ubuntu-1.19.10+1 ✅  Done
```

PR #787, opened 2024-04-08
Change revision type

Itxaka:
Otherwise the bumper always things there is a new version


jimmykarily:
This PR is only replacing `+` with `-`. How is `-` better and why does it fix anything? I think it's most common to split the build number in versions with a `+`, no?

Itxaka:
> This PR is only replacing `+` with `-`. How is `-` better and why does it fix anything? I think it's most common to split the build number in versions with a `+`, no?

Yep yep, that's what I thought but the autobumpers does not think so and thinks that 255 is higher than 255+4 

In other cases with the revision with a dash, it seems to work as expected so I though this might work.

jimmykarily:
Oh! gotcha. Maybe it thinks `+N` is a pre-release semver-wise: https://semver.org/#spec-item-10

Go ahead then. Though we still use `+` in other packages I think. We should try to be consistent.

PR #811, opened 2024-04-17
Re-introduce webui dependencies for kairos-agent

mauromorales:
These were removed by mistake when making the binaries smaller for UKI

relates to kairos-io/kairos#2473


mudler (approved):

Itxaka (approved):
oooh, nice catch!

PR #843, opened 2024-05-07
Fix webui for agent

Itxaka:
It was mistakenly added to the env steps but it should be run in the steps instead

Fixes: https://github.com/kairos-io/kairos/issues/2520

### DIFF
--- a/packages/system/kairos-agent/collection.yaml
+++ b/packages/system/kairos-agent/collection.yaml
@@ -12,7 +12,7 @@ packages:
     description: "Lifecycle agent for kairos"
   - name: "kairos-agent"
     category: "fips"
-    version: "2.10.2+1"
+    version: "2.10.3"
     labels:
       github.repo: "kairos-agent"
       autobump.revdeps: "true"
--- a/packages/system/kairos-agent/collection.yaml
+++ b/packages/system/kairos-agent/collection.yaml
@@ -1,7 +1,7 @@
 packages:
   - name: "kairos-agent"
     category: "system"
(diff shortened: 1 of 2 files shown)


mauromorales (approved):

packages/system/kairos-agent/collection.yaml
```diff
@@ -1,7 +1,7 @@
 packages:
   - name: "kairos-agent"
     category: "system"
-    version: "2.10.2"
+    version: "2.10.2+1"
```
mauromorales, line 4:
```suggestion
    version: "2.10.3"
```

packages/system/kairos-agent/collection.yaml
```diff
@@ -12,7 +12,7 @@ packages:
     description: "Lifecycle agent for kairos"
   - name: "kairos-agent"
     category: "fips"
-    version: "2.10.2"
+    version: "2.10.2+1"
```
mauromorales, line 15:
```suggestion
    version: "2.10.3"
```

mauromorales:
let's use this PR to also bump the agent https://github.com/kairos-io/packages/pull/844

packages/system/kairos-agent/collection.yaml
Itxaka: not really no? version is the fixed one from agent, this is a revision of the package only, nothing changed upstream in the agent

Itxaka: crap, didnt see your comment below loool

PR #864, opened 2024-05-23
add well known persistent mounts

santhoshdaivajna:
Rook-Ceph
-     /var/lib/rook
Portworx

-    /etc/pwx
-    /var/lib/osd
-    /var/cores

Tailscale:

-    /var/lib/tailscale

Longhorn

-    /var/lib/longhorn

Cilium
- /var/run/cilium


Itxaka (approved):
look ok but you would need to bump the version of the package under packages/static/kairos-overlay-files/collection.yaml so it rebuilds it

PR #943, opened 2024-07-04
Add fips go1.22.5

jimmykarily:
part of: https://github.com/kairos-io/kairos/issues/2670

### DIFF
--- a/packages/toolchain-go/collection.yaml
+++ b/packages/toolchain-go/collection.yaml
@@ -68,12 +68,18 @@ packages:
   - name: toolchain-go
     category: fips
     variant: "alpine"
-    version: "1.19.10"
+    version: "1.22.5"
+    tag: "1.22.5"
+    hidden: true
+  - name: toolchain-go-ubuntu
+    category: fips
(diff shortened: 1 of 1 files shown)


mauromorales (approved):
LGTM

packages/toolchain-go/collection.yaml
```diff
@@ -68,12 +68,18 @@ packages:
   - name: toolchain-go
     category: fips
     variant: "alpine"
-    version: "1.19.10"
-    tag: "1.19.10"
+    version: "1.22.5"
```
Itxaka, line 71: why we are moving alpine from 1.19 to 1.22 but on bullseye we are adding an extra package?

jimmykarily: we are adding an extra package in both cases. But in the diff it look like we are "moving" the alpine one, because in parallel, I'm removing the duplicate entry for 1.19.10 :D.

jimmykarily: it's clearer if you use the "split" view

PR #944, opened 2024-07-04
use values version for the proper version build

Itxaka:
otherwise we are using an old version with a newer package tag, which is wrong

### DIFF
--- a/packages/utils/kubectl/build.yaml
+++ b/packages/utils/kubectl/build.yaml
@@ -5,7 +5,7 @@ requires:
 
 package_dir: /kubectl
 env:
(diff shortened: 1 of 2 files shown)


jimmykarily:
What is the `tag` used for? Should it be removed? If not, shouldn't it match the `version`?

Itxaka:
> What is the `tag` used for? Should it be removed? If not, shouldn't it match the `version`?

now its removed as the original didnt work as expected.

Not sure why was the reason for this but it measn that our 1.30.x package actually ships kubectl 1.29.x xD

jimmykarily (approved):

PR #945, opened 2024-07-04
Fix autobumper sed

Itxaka:


packages/k8s/k3s/collection.yaml
```diff
@@ -6,7 +6,7 @@ packages:
     labels:
       github.owner: "k3s-io"
       github.repo: "k3s"
-      autobump.sed_script: 's/\+k3s1//g'
+      autobump.sed_script: 's/\+k3s1\|+k3s2//g'
```
mauromorales, line 9: I guess it can be something like k3s1[1-9]

Itxaka: Yeah I have no idea what it means lol

mauromorales: is how they do a patch release, because they pin their releases to kubernetes releases and never diverge from them

jimmykarily (approved):

packages/k8s/k3s/collection.yaml
jimmykarily: We hardcode it here though: https://github.com/kairos-io/kairos/blob/6d68929655235cec5f752f8b2f2cb12553aded94/Earthfile#L242-L251

Itxaka: umm, good catch. Does that mean that we only accept versions with the suffix? our upstream versions do not have that suffix? https://packages.kairos.io/Kairos/k8s/k3s-systemd/1.30.2/

Reading the TODO seems that we already fixed that and could be dropped no?

Itxaka: anyway, updated with @mauromorales suggestion, much nicer in the long run :D

Itxaka:
wait wtf why it merged? shiit we dont have the tests be required for merge craaap.

PR #970, opened 2024-07-15
set kairos services to simple

Itxaka:
If they are of type oneshot, systemd will wait until the pid exits before markling the service as running. Instead it will mark them as starting which is wrong and on livecd it causes the systemctl status output to be on starting forever.

Instead lets set those services to simple, which means they will be marked as started once the binary has been launched.

reset makes sense to have them like that as we dont want anything else to continue, but our reset and reboot after that.

Fixes https://github.com/kairos-io/kairos/issues/2665

### DIFF
--- a/packages/static/kairos-overlay-files/collection.yaml
+++ b/packages/static/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.1.41"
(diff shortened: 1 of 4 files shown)


Itxaka:
looks like the target is also wrong, as its sysinit instead of multiuser target?

Testing this

Itxaka:
kairos service works as expected

recovery service works as expected

interactive installer works as expected

PR #977, opened 2024-07-18
Build stub with tpm support :facepalm:

Itxaka:
According to the compile options this enables tpm to be extended....

```
option('tpm', type : 'boolean',
       description : 'TPM should be used to log events and extend the registers')
```


Itxaka:
manually tested and it works

PR #978, opened 2024-07-22
Bump systemd-boot/stub to 256

Itxaka:
Fixes https://github.com/kairos-io/kairos/issues/2728


mauromorales (approved):
do we need to wait until we are sure there's no 2.1.2?

Itxaka:
> do we need to wait until we are sure there's no 2.1.2?

good point. Too late now :D If needed we can revert this PR I guess?

PR #988, opened 2024-07-30
Generate profile config instead of having a static file

mauromorales:
The file `/etc/bash.bashrc.local` is only read by SUSE, the other distros (including SUSE) prefer a file under `/etc/profile.d/`. It's also generating the file instead of having it static in case one of the distros needs it differently then we can use flow control

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/32_profile.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/32_profile.yaml
@@ -1,0 +1,15 @@
+name: "Profile extensions"
+stages:
+  initramfs:
+    - name: "Set KUBECONFIG to access k9s and kubectl"
+      files:
+        - path: /etc/profile.d/kairos.sh
+          permissions: 0644
+          owner: 0
+          group: 0
+          content: |
+            if [ -z "$KUBECONFIG" ]; then
+                if [ -e /etc/rancher/k3s/k3s.yaml ]; then
+                    export KUBECONFIG="/etc/rancher/k3s/k3s.yaml"
+                fi
+            fi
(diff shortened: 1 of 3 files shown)


packages/static/kairos-overlay-files/files/system/oem/32_profile.yaml
```diff
@@ -0,0 +1,15 @@
...
+  initramfs:
+    - name: "Set KUBECONFIG to access k9s and kubectl"
+      files:
+        - path: /etc/profile.d/kairos.sh
+          permissions: 0644
+          owner: 0
+          group: 0
+          content: |
+            if [ -z "$KUBECONFIG" ]; then
+                if [ -e /etc/rancher/k3s/k3s.yaml ]; then
```
jimmykarily, line 12: I think the first time the system boots, it takes a while for k3s to create this file. This is running in the `initrafms` stage so I'm not sure if the file would be there at that point.

packages/static/kairos-overlay-files/files/system/oem/32_profile.yaml
```diff
@@ -0,0 +1,15 @@
+name: "Profile extensions"
+stages:
+  initramfs:
+    - name: "Set KUBECONFIG to access k9s and kubectl"
+      files:
+        - path: /etc/profile.d/kairos.sh
```
jimmykarily, line 6: do all scripts under this path run automatically on login?

packages/static/kairos-overlay-files/files/system/oem/32_profile.yaml
jimmykarily: wait, my mistake. This is not running at that point, it's only been created. Ignore my comment.

jimmykarily (approved):

packages/static/kairos-overlay-files/files/system/oem/32_profile.yaml
jimmykarily: I just read the PR description. All good.

PR #1003, opened 2024-08-05
Re-enable install section 

Itxaka:


jimmykarily (approved):

jimmykarily:
So the reason it's not going to start all the services again (reset, interactive etc) is because we now have an empty `/etc/machine-id` file right?

Itxaka:
> So the reason it's not going to start all the services again (reset, interactive etc) is because we now have an empty `/etc/machine-id` file right?

Basically yes. Otherwise services are stuck on boot

PR #1011, opened 2024-08-14
Fix home directory ownership

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/2797

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/00_home_dir_owner_fix.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/00_home_dir_owner_fix.yaml
@@ -1,0 +1,15 @@
+name: "Fix home directory permissions (kairos issue #2797)"
+stages:
+  initramfs.after:
+    - name: "Fix permissions"
+      commands:
+        - |
+          # Iterate over users in /etc/passwd and chown their directories
+          awk -F: '$3 >= 1000 && $6 ~ /^\/home\// {print $1, $6}' /etc/passwd | while read -r user homedir; do
+              if [ -d "$homedir" ]; then  # Check if the home directory exists
+                  echo "Changing ownership of $homedir to $user"
+                  chown -R "$user":"$user" "$homedir"
+              else
+                  echo "Directory $homedir does not exist for user $user"
+              fi
+          done
--- a/packages/static/kairos-overlay-files/collection.yaml
+++ b/packages/static/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.1.45"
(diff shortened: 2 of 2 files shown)


packages/static/kairos-overlay-files/files/system/oem/00_home_dir_owner_fix.yaml
```diff
@@ -0,0 +1,15 @@
+name: "Fix home directory permissions (kairos issue #2797)"
+stages:
+  initramfs.after:
```
jimmykarily, line 3: I wonder if we should somehow make it an one-off fix. This will now run on every boot. With a sentinel file or something. Or maybe we should add a check and only do the `chown` if the owner is not the expected one?

jimmykarily: ok adapted the script to do that.

jimmykarily: nope. This doesn't work. The home directory itself was already owned by `kairos`. It's those files inside that were wrong (e.g. `authorized_keys` in the `.ssh` directory). I'll revert.

packages/static/kairos-overlay-files/files/system/oem/00_home_dir_owner_fix.yaml
```diff
@@ -0,0 +1,15 @@
...
+    - name: "Fix permissions"
+      commands:
+        - |
+          # Iterate over users in /etc/passwd and chown their directories
+          awk -F: '$3 >= 1000 && $6 ~ /^\/home\// {print $1, $6}' /etc/passwd | while read -r user homedir; do
+              if [ -d "$homedir" ]; then  # Check if the home directory exists
+                  echo "Changing ownership of $homedir to $user"
+                  chown -R "$user":"$user" "$homedir"
```
Itxaka, line 11: `uid $user -i` to get the user id, same for the group? so we set the user:group to numbers instead of names?

jimmykarily: Does it make any difference? I think it doesn't matter how you refer to the user and group, in the end it's going to set the ownership the same way. E.g.

```
~ $ sudo chown 1000:1001 deleteme
~ $ ls -liah deleteme
9717519 -rw-r--r-- 1 dimitris dimitris 0 Αυγ  14 14:32 deleteme
~ $ sudo chown dimitris:dimitris deleteme
~ $ ls -liah deleteme
9717519 -rw-r--r-- 1 dimitris dimitris 0 Αυγ  14 14:32 deleteme
```

mauromorales (approved):

PR #1022, opened 2024-08-27
Disable selinux for all flavors

jimmykarily:
because in order to support selinux, we need to do more than just enable it in the cmdline.

Fixes https://github.com/kairos-io/kairos/issues/2802

### DIFF
--- a/packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
+++ b/packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
@@ -1,14 +1,18 @@
 function setSelinux {
     source (loop0)/etc/os-release
-    set baseSelinuxCmd=""
-    if test $KAIROS_FAMILY == "rhel" -o test $ID == "opensuse-tumbleweed" -o test $ID == "opensuse-leap"; then
-        set baseSelinuxCmd="selinux=0"
-    else
-        # if not in recovery
-        if [ -z "$recoverylabel" ];then
-            set baseSelinuxCmd="security=selinux selinux=1"
-        fi
-    fi
+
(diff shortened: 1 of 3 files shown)


packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
```diff
@@ -1,14 +1,18 @@
...
-        if [ -z "$recoverylabel" ];then
-            set baseSelinuxCmd="security=selinux selinux=1"
-        fi
-    fi
+
+    # Disable selinux for all distros. Supporting selinux requires more than
+    # just enabling it like this.
+    set baseSelinuxCmd="selinux=0"
+
+    #if test $KAIROS_FAMILY == "rhel" -o test $ID == "opensuse-tumbleweed" -o test $ID == "opensuse-leap"; then
```
mauromorales, line 8: since this is applied to all other distros, could it be that it will break any of the boards? since those are not tested, thinking mainly about nvidia 🤷

mauromorales (approved):
I don't see any issues with the code but I wonder what impact this will have since we cannot really test easily on all different cases e.g. boards, auroraboot, different flavors, different features. But maybe we can merge it and at least have it on master and then we start seeing the tests

packages/static/kairos-overlay-files/files/etc/cos/bootargs.cfg
jimmykarily: I can't think of a reason why a board would require selinux to work :shrug:

jimmykarily:
The tests passed on the other (hackish) PR. I will merge this.

PR #1024, opened 2024-08-30
Increase UDP buffer sizes

mauromorales:
The new size is 7.5M as recommended by quic-go [1] which previously was 2.5M [2]

[1]: https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes/a3327deff89d2428d48596ce0e643531f9944f99
[2]: https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes/597639d834d5d6c242d37d49a02ed04ca65332c9

Also I'm not sure if these should also be increased, or removed, or kept as they are. For edgevpn it is needed, but why for the rest?

https://github.com/kairos-io/packages/blob/9a729607cf29ae8036e2e377e4676cfbaa9fd5cf/packages/static/kairos-overlay-files/files/system/oem/09_systemd_services.yaml#L10

https://github.com/kairos-io/packages/blob/9a729607cf29ae8036e2e377e4676cfbaa9fd5cf/packages/static/kairos-overlay-files/files/etc/systemd/system/kairos-recovery.service#L10

### DIFF
--- a/packages/utils/edgevpn-systemd/edgevpn.service
+++ b/packages/utils/edgevpn-systemd/edgevpn.service
@@ -5,7 +5,8 @@ After=network.target
 [Service]
 EnvironmentFile=/etc/systemd/system.conf.d/edgevpn.env
 LimitNOFILE=49152
-ExecStartPre=-/bin/sh -c "sysctl -w net.core.rmem_max=2500000"
+ExecStartPre=-/bin/sh -c "sysctl -w net.core.rmem_max=7500000"
+ExecStartPre=-/bin/sh -c "sysctl -w net.core.wmem_max=7500000"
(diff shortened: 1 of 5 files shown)


packages/utils/edgevpn-systemd/edgevpn.service
```diff
@@ -5,7 +5,8 @@ After=network.target
 [Service]
 EnvironmentFile=/etc/systemd/system.conf.d/edgevpn.env
 LimitNOFILE=49152
-ExecStartPre=-/bin/sh -c "sysctl -w net.core.rmem_max=2500000"
```
jimmykarily, line 8: Let's add the link to where we found this number

jimmykarily (approved):

PR #1025, opened 2024-09-03
Create a sentinel file to avoid running the fix every time

jimmykarily:
This file can also be created by the user to skip this fix if they so desire. For example, they may have files owned by some other user and they don't want to mess it up. In that case they can run a similar script themselves, but with additional logic.

Fixes https://github.com/kairos-io/kairos/issues/2843


jimmykarily:
I tested manually and works as expected

Itxaka (approved):

mauromorales (approved):

PR #1030, opened 2024-09-04
Add override for the tpm2 service

jimmykarily:
to avoid waiting for 90 seconds for the tpm device to appear when there is not tpm device

Fixes https://github.com/kairos-io/kairos/issues/2835

### DIFF
--- a/packages/static/kairos-overlay-files/files/etc/systemd/system/tpm2-abrmd.service.d/override.conf
+++ b/packages/static/kairos-overlay-files/files/etc/systemd/system/tpm2-abrmd.service.d/override.conf
@@ -1,0 +1,6 @@
+# Don't wait for 90 seconds for tpm devices when they don't exist
+[Service]
+TimeoutSec=5
+
+[Unit]
+JobTimeoutSec=5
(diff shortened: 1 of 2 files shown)


packages/static/kairos-overlay-files/files/etc/systemd/system/tpm2-abrmd.service.d/override.conf
```diff
@@ -0,0 +1,3 @@
+[Service]
+# Don't wait for 90 seconds for tpm devices when they don't exist
+TimeoutStopSec=5
```
jimmykarily, line 3:
```suggestion
TimeoutStartSec=5
```

Itxaka (approved):

Itxaka:
We sure then that no other jobs will need the same timeout?

jimmykarily:
> We sure then that no other jobs will need the same timeout?

I checked with `systemctl list-dependencies` and such and that's the only service relying on `dev-tpm0.device` . Also used `grep -R` to check in unit files to be sure. Tests also show that there is no delay so :shrug: .

jimmykarily:
In opensuse the `tpm2-abrmd` service is not enabled by default: 

image

I can start it manually though just fine. Also there is no delay while booting (after all it's not enabled?). All good?

PR #1036, opened 2024-09-09
Fix home dir ownership after boot

jimmykarily:
because users might be created in various stages (like "network")

### DIFF
--- a/packages/static/kairos-overlay-files/files/usr/bin/fix-home-dir-ownership
+++ b/packages/static/kairos-overlay-files/files/usr/bin/fix-home-dir-ownership
@@ -1,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+SENTINEL_FILE="/usr/local/.kairos/skip-home-directory-ownership-fix"
+
+if [ -f $SENTINEL_FILE ]; then
+    echo "Skipping ownership fix because sentinel file was found: $SENTINEL_FILE"
+    exit 0
+fi
+
+# Iterate over users in /etc/passwd and chown their directories
+awk -F: '$3 >= 1000 && $6 ~ /^\/home\// {print $1, $6}' /etc/passwd | while read -r user homedir; do
+    if [ -d "$homedir" ]; then  # Check if the home directory exists
+        echo "Changing ownership of $homedir to $user"
+        chown -R "$user":"$user" "$homedir"
+    else
+        echo "Directory $homedir does not exist for user $user"
+    fi
+done
+
+# Write the sentinel file
+mkdir -p "$(dirname $SENTINEL_FILE)"
+echo "https://github.com/kairos-io/kairos/issues/2843" > $SENTINEL_FILE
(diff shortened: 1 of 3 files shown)


packages/static/kairos-overlay-files/files/usr/bin/fix-home-dir-ownership
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+
+SENTINEL_FILE="/usr/local/.kairos/skip-home-directory-ownership-fix"
```
jimmykarily, line 3: I moved the whole logic in a script to allow users to run this manually without having to copy-paste the whole script and also to be able to easily run it in other stages as well if needed.

@Itxaka with a config like this:

```
#cloud-config

install:
    poweroff: false
    reboot: false
stages:
    network:
        - name: "Setup users"
          users:
             dimitris:
               passwd: "kairos"
users:
    - name: kairos
      passwd: kairos
```

the second user (`dimitris`) will be created in the `network` stage. But in a similar manner, a user might be created in any stage, right? Is there a stage we can run the script in  that is guaranteed to always run after users are created? Otherwise we should maybe run in in `boot.after` and document in the release notes that if they users need the fix in different stages, they should add it to their cloud config file. What do you think?

Itxaka: yeah they can be created in any stage so we should go to the last step that we know off (network?) and run it there in network.after to be sure.

Or... can we add the fix into yip directly? I.e. in the user creation we add a chown just after creating the user+sentinel to run it just once? After all, the fix was done on yip so it should be a yip thing to fix any shortcomings? maybe the sentinel cant be done on that part?

jimmykarily: It would be good if we could stop applying this fix after some version because it could potentially break the ownership of some files in the home dir. It's the best we could do as a fix for what we introduced but imagine someone installing a Kairos for the first time, using a fixed version and having to deal with messed up permissions introduced to fix problem that is 2 years old (which he never encountered).

If our suggestion to users was to upgrade to all intermediate versions, we could even remove the fix in the next version (because it only has to be applied once). But I don't think it's our recommendation. After all, upgrades are atomic and usually it's no problem to skip a version.

Of course little difference does it make if the fix is in yip or in a cloud config file, although with time, reverting a fix in code might be more tricky that just deleting a couple of scripts.

jimmykarily: I used `network.after` and it works with the provided config here. In the release notes we can document the existence of that script and a warning that if people are creating users in other stages, they should make sure script is running after their users are created. We will need to document the existence of the sentinel file anyway, for users that want to skip the fix.

Itxaka (approved):
good to me!

mauromorales:
Fixes kairos-io/kairos#2797

PR #1047, opened 2024-09-13
Build arm in our new arm64 worker

Itxaka:


packages/system/kairos-agent/collection.yaml
```diff
@@ -1,7 +1,7 @@
 packages:
   - name: "kairos-agent"
     category: "system"
-    version: "2.13.7"
+    version: "2.14.0"
```
jimmykarily, line 4: there is no such version of kairos-agent (?)

Itxaka: there is, but it should be an rc as I screwed up :( https://github.com/kairos-io/kairos-agent/releases/tag/v2.14.0

Itxaka:
seems like this cuts the time it takes to build for arm to half which is good

jimmykarily (approved):

PR #1059, opened 2024-09-19
Add `#cloud-config` header to all overlay files

jimmykarily:
because `kairos-agent config` command ignores those files because of the missing header. That command is supposed to print all the config that is taken into account from kairos-agent and immucore and all.

Currently it ignore those files the same way it ignores them when running `kairos-agent manual-install` et al. Adding the header will make the kairos-agent respect those files, essentially with no effect since none of these files set any installation options.
The only useful result will be that `kairos-agent config` will print the full config making it easier for the user to see the full merged yaml that results from all the available config files.

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
@@ -1,3 +1,5 @@
+#cloud-config
+
 name: "Datasource handling"
 stages:
   rootfs.before:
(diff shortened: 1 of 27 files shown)


Itxaka (changes requested):
Against this. This are system config files, not agent action configs. To me there is a clear disctinction:

 - action configs affect the install/upgrade/reset actions. They can run system configurations as well
 - system config files configure the system

we have to think, whats the `kairos-agent config` suposed to show? the config for install/reset/upgrade? Or the full system configuration?

jimmykarily:
We'll do this instead: https://github.com/kairos-io/packages/pull/1060

We will remove that directory from the directories scanned by the kairos-agent and we'll document it. That directory should only be used for stage definitions.

PR #1063, opened 2024-09-23
Add rpi kernel

Itxaka:
Builts from upstream directly. It would require changes to the arm builder scripts to identify the source as ubuntu and use this package instead, plus copy the /lib provided into the base image.

 - remove upstream kernel from base image
 - remove /lib/modules/* from base image
 - copy kernel to boot and links to to /boot/vmlinuz
 - copy package /lib into base image

supposedly that should make ubuntu boot back as the kernel has EFI enabled by default but no idea if it would actually work.

We could also split the package into modules+kernel and dtb files and overlay but I think its useless now.

### DIFF
--- a/packages/kernels/linux-rpi/build.yaml
+++ b/packages/kernels/linux-rpi/build.yaml
@@ -1,0 +1,17 @@
+package_dir: /out
+
+image: ubuntu:24.04
+
+prelude:
+  - apt-get update -y && apt-get install -y gcc gcc-aarch64-linux-gnu git make flex bc bison libssl-dev xz-utils kmod
+  - git clone --branch {{ .Values.version }} --depth 1 https://github.com/raspberrypi/linux.git linux
+
+steps:
+  - mkdir -p make /build
+  - mkdir -p /out/boot/overlays
+  - cd linux && CROSS_COMPILE=aarch64-linux-gnu- ARCH=arm64 make O=/build bcm2711_defconfig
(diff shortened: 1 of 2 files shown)


packages/kernels/linux-rpi/build.yaml
```diff
@@ -0,0 +1,17 @@
+package_dir: /out
+
+image: ubuntu:24.04
+
+prelude:
+  - apt-get update -y && apt-get install -y gcc gcc-aarch64-linux-gnu git make flex bc bison libssl-dev xz-utils kmod
+  - git clone --branch {{ .Values.version }} --depth 1 https://github.com/raspberrypi/linux.git linux
```
jimmykarily, line 7: This pulls their default branch (`rpi-6.6.y` ?). Should we make it a value to avoid surprises? (so that we can manually bump)

jimmykarily: I'm stupid. You already do :facepalm:

jimmykarily (approved):

packages/kernels/linux-rpi/collection.yaml
```diff
@@ -0,0 +1,9 @@
+packages:
+  - name: "linux-rpi"
+    category: "kernels"
+    version: "rpi-6.6.y"
```
mauromorales, line 4: what's the y for?

mauromorales (approved):

packages/kernels/linux-rpi/collection.yaml
Itxaka: thats what they call their branch names lol . Instad of 6.6.x they use the y for some reason :D

Maybe they did rpi-6.x.y before?

PR #1064, opened 2024-09-23
Build the latest 3 k3s versions

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/2878

### DIFF
--- a/packages/k8s/k3s/build.yaml
+++ b/packages/k8s/k3s/build.yaml
@@ -10,7 +10,7 @@ env:
   - INSTALL_K3S_SKIP_SELINUX_RPM="true"
   # Remove any possible "+N" from the end of the version. We only bump that to
   # avoid overwritting existing images when we want the package to be rebuilt.
-  - INSTALL_K3S_VERSION=v{{ regexReplaceAll "\\+\\d+$" .Values.version "" }}+k3s1
+  - INSTALL_K3S_VERSION=v{{ regexReplaceAll "\\+\\d+$" .Values.version "" }}+k3s{{.Values.k3s_version}}
   {{$arch:=(default "amd64" .Values.arch)}}
   {{ if eq $arch "arm" }}
   - ARCH=arm64
--- a/packages/k8s/k3s/collection.yaml
+++ b/packages/k8s/k3s/collection.yaml
@@ -2,7 +2,7 @@ packages:
   - name: k3s-openrc
     category: k8s
     version: "1.31.1"
-    k3s_version: "2"
+    k3s_version: "1"
     labels:
       github.owner: "k3s-io"
       github.repo: "k3s"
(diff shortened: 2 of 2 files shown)


packages/k8s/k3s/build.yaml
```diff
@@ -10,7 +10,7 @@ env:
   - INSTALL_K3S_SKIP_SELINUX_RPM="true"
   # Remove any possible "+N" from the end of the version. We only bump that to
   # avoid overwritting existing images when we want the package to be rebuilt.
-  - INSTALL_K3S_VERSION=v{{ regexReplaceAll "\\+\\d+$" .Values.version "" }}+k3s1
+  - INSTALL_K3S_VERSION=v{{ regexReplaceAll "\\+\\d+$" .Values.version "" }}+k3s{{.Values.k3s_version}}
```
jimmykarily, line 13: I don't know why we removed this. Looking at `git blame`, we used to have it and that's the reason why `k3s_version` is still around in the collection.yaml below.

Itxaka: I think it is because the updater will update the version but not the revision. So it would use a version but report another?

jimmykarily: but the updater is already doing a lousy job updating only one of them. If we are manually bumping, then we can as well keep `k3s_version` around since every now and then [a `-k3s2` may appear](https://github.com/k3s-io/k3s/releases/tag/v1.30.2%2Bk3s2).

@mauromorales it's not always `k3s1`, see my link above.

jimmykarily:
> Should probably make the auto update not trigger this anymore?

Where is that setting? Is it just the `autobump.sed_script` settings in collection.yaml that trigger it?

packages/k8s/k3s/build.yaml
mauromorales: Yup I'm aware it's not always 1. If we don't plan to use the auto update at all, then it's ok. Maybe I misunderstood the PR

mauromorales:
I think as long as it is in the directory it will try to auto bump. But don't remember correctly

packages/k8s/k3s/build.yaml
jimmykarily: I don't see a way to exclude a dir: https://github.com/Luet-lab/extensions/blob/master/extensions/autobump-github/luet-autobump-github

We can ignore the bumping PR maybe?

Itxaka:
I thnk that the problem is that the autobumper will still try to bump it anyway and the automerge will kick in and break the version again?

jimmykarily:
> I thnk that the problem is that the autobumper will still try to bump it anyway and the automerge will kick in and break the version again?

We blindly approve all PRs? https://github.com/kairos-io/packages/blob/ae54338fa8fb13f42019bfae1103c4d109baf47d/.github/workflows/autoapprove.yml#L20-L25

:D

PR #1070, opened 2024-09-24
Drop tpm2 override

Itxaka:
Problem was that we shipped a wrong userspace daemon for tpm2 and we didnt need it

With this merged, we should not need this workaround anymore: https://github.com/kairos-io/kairos/pull/2884


jimmykarily (approved):

jimmykarily:
Nice finding!

mauromorales (approved):
cool cool cool

PR #1081, opened 2024-09-28
Fix online status with more than one NIC

bencorrado:
For systems with more than one unconfigured NIC, this allows the system to see any one interface online as enough to proceed with online status. Fixes [kairos:#2898](https://github.com/kairos-io/kairos/issues/2898)

### DIFF
--- a/packages/static/kairos-overlay-files/collection.yaml
+++ b/packages/static/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.1.56"
+    version: "1.1.57"
--- a/packages/static/kairos-overlay-files/collection.yaml
+++ b/packages/static/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.1.53"
+    version: "1.1.56"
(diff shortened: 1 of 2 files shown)


Itxaka (changes requested):
I would drop the `--ipv4` in case we got a netwokr with only ipv6 maybe? if that is not specified, it seems that it will use the configuration from the device to know which version to use so it may be better that way.

Also this will fail to apss CI becuase you need to bump the minor version of the package, so it gets properly build. Just bump the patch version on https://github.com/kairos-io/packages/blob/main/packages/static/kairos-overlay-files/collection.yaml#L4 :)

Itxaka (approved):

Itxaka:
hey @bencorrado sorry about this, updating the branch made the version go away as we already bumped it in a different Pr. Could you bump the version? Will merge it afterwards

Itxaka:
oh nice, I can push it as well :D

bencorrado:
Thanks for merging and shepherding me through @Itxaka 

Did this count for #hacktoberfest?  I want a t-shirt!

mauromorales:
@bencorrado I just added the tag, if you are subscribed to hacktoberfest, I would hope that it will eventually show up as counting on their app. As for the kairos tshirt you're in, I'm adding you to the list https://github.com/kairos-io/kairos/discussions/2924

PR #1087, opened 2024-10-02
Implement `--recovery` flag in suc-upgrade script

jimmykarily:
To allow users to upgrade recovery from kubernetes

Part of: https://github.com/kairos-io/kairos/issues/2760


mauromorales (approved):
noicceee

PR #1088, opened 2024-10-02
Bring alpine initrd closer to other flavors

Itxaka:
Set the dirs to be the same as other flavors, especially the livecd one as we read the configs from that dir, so kairos tooling expects the livecd to be mounted there

Fixes: https://github.com/kairos-io/kairos/issues/2912

### DIFF
--- a/packages/alpine/files/initramfs-init
+++ b/packages/alpine/files/initramfs-init
@@ -1,6 +1,6 @@
 #!/bin/sh
 
-VERSION=3.8.1-kairos3
+VERSION=3.8.1-kairos4
 INIT=/sbin/init
 sysroot="$ROOT"/sysroot
 # This is where we mount the livecd
@@ -209,9 +209,9 @@ retry() {
   done
 }
 
-# uses the first "eth" interface with state 'up'.
+# uses the first network interface with state 'up'.
 ip_choose_if() {
-	for x in "$ROOT"/sys/class/net/eth*; do
+	for x in "$ROOT"/sys/class/net/*; do
 		if grep -iq up $x/operstate 2>/dev/null;then
 			[ -e "$x" ] && echo ${x##*/} && return
 		fi
--- a/packages/alpine/files/initramfs-init
+++ b/packages/alpine/files/initramfs-init
@@ -1,8 +1,16 @@
 #!/bin/sh
 
-VERSION=3.8.1-kairos2
+VERSION=3.8.1-kairos3
 INIT=/sbin/init
 sysroot="$ROOT"/sysroot
+# This is where we mount the livecd
+liveroot=/run/initramfs/live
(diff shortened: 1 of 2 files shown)


packages/alpine/files/initramfs-init
```diff
@@ -375,16 +383,16 @@ if grep -q netboot /proc/cmdline; then
   sync
   # Create mountpoints
   ebegin "Create mountpoints"
-  mkdir -p /media/root-rw /run/rootfsbase $sysroot/media/root-rw $sysroot/run/rootfsbase
+  mkdir -p $rootRW $rootfsbase $sysroot$rootRW $sysroot$rootfsbase
```
mauromorales, line 386: if I understand correctly this:

```
/media/root-rw /run/rootfsbase $sysroot/media/root-rw $sysroot/run/rootfsbase
```

Becomes this

```
/run/overlay /run/rootfsbase $sysroot/run/overlay $sysroot/run/rootfsbase
```

so that means that `/media/root-rw` got changed to `/run/overlay` and same to the one that is prefixed by the `$sysroot`

why did the location change?

mauromorales (approved):
smart approach changing them to variables 👍

packages/alpine/files/initramfs-init
Itxaka: to have it in line with what immucore does. We mount a writable /run/overlay and then submounts things there (/etc /var and so on, basically the ephemeral parts)

So in this case, this initrd was just doing its own thing. In this case it should not matter much as nothing should be using that, especially during live boot, but having ti be the same makes it easier to understand both what dracut does and what we do :)

packages/alpine/files/initramfs-init
```diff
@@ -398,20 +406,20 @@ if grep -q cdroot /proc/cmdline ;then
   label=$(grep -o CDLABEL.* /proc/cmdline | cut -f 1 -d ' ' | cut -f 2 -d '=' )
   # Create mountpoints
   ebegin "Create mountpoints"
-  mkdir -p /media/root-ro /media/root-rw /run/rootfsbase $sysroot/media/root-ro $sysroot/media/root-rw $sysroot/run/rootfsbase
+  mkdir -p $liveroot $rootRW $rootfsbase $sysroot$liveroot $sysroot$rootRW $sysroot$rootfsbase
```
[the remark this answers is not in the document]
Itxaka: same thing. On dracut it automounts the livecd under that path but we were mounting it under root-ro because the upstream init from alpine did that. It makes no sense for us to deviate from what we expect and it shows, this was causing alpine to not autoinstall with a config bundled, because the livecd was mounted under that path and agent looks under /run/initramfs/live to read any configs coming from the live media.

PR #1096, opened 2024-10-04
fix(k9s): use version attribute for builds

thomasthep:
Build was using tag and version bump wasn't happening. With this change, we revert back to using the version attribute and remove tag as this is not necessary and would require syncing tags and versions. Furthermore, the version now removes all '+' instead of '-'.


Itxaka (approved):

Itxaka:
Good catch!

PR #1111, opened 2024-10-14
Add yaml linter

Itxaka:
Fixes https://github.com/kairos-io/kairos/issues/2920


Itxaka:
Validate is gonna fail because there are changes to the files but they werent bumped. But I dont think we need to really bump them for this as its mainly cleanup

mauromorales (approved):
I thought a lot more issues would arise, nice to see it's not the case!

Itxaka:
dont want to merge it now, as there are a few changes to yaml files that should be gone first in case this breaks stuff

PR #1137, opened 2024-11-11
Pull data sources at install

bencorrado:
Improve the way we pull in datasources.

### DIFF
--- a/packages/utils/edgevpn/definition.yaml
+++ b/packages/utils/edgevpn/definition.yaml
@@ -1,6 +1,6 @@
 name: edgevpn
 category: utils
-version: "0.28.3+1"
+version: "0.28.3-1"
 labels:
   github.repo: "edgevpn"
   github.owner: "mudler"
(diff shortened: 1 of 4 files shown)


packages/utils/edgevpn/definition.yaml
```diff
@@ -1,7 +1,6 @@
 name: edgevpn
 category: utils
-version: "0.28.3"
-tag: "0.24.6"
```
jimmykarily, line 4: We should keep the `version` one, because the pipeline that builds the packages when the version is bumped, relies on this field. The change should be minimal. Let's just change this to `.Values.version`: https://github.com/kairos-io/packages/blob/608b1d23a802dbe902d0a5f8cdf722d393bd255c/packages/utils/edgevpn/build.yaml#L12

jimmykarily (approved):
Other than my comment, the rest looks great :). Thanks @bencorrado!

bencorrado:
> Other than my comment, the rest looks great :). Thanks @bencorrado!

You should have permissions to add to the pr, if you have time to make that change, you don't need to wait for me, I am already in bed for the night.

jimmykarily (approved):

PR #1138, opened 2024-11-13
Add auroraboot

jimmykarily:
Part of: https://github.com/kairos-io/kairos/issues/1633

(to have it as an installation option in cases where we used to install enki [like this](https://github.com/kairos-io/osbuilder/blob/3aa1fa1104d099c382a09325ff85fe1c21cfeb47/tools-image/Dockerfile#L106))

### DIFF
--- a/packages/toolchain-go/collection.yaml
+++ b/packages/toolchain-go/collection.yaml
@@ -2,7 +2,7 @@ packages:
   - name: toolchain-go
     category: development
     variant: "alpine"
-    version: "1.23.2"
+    version: "1.23.3"
     hidden: true
     labels:
       autobump.revdeps: "false"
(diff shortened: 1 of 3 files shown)


packages/toolchain-go/collection.yaml
```diff
@@ -2,7 +2,7 @@ packages:
   - name: toolchain-go
     category: development
     variant: "alpine"
-    version: "1.23.2"
+    version: "1.23.3"
```
jimmykarily, line 5: the auroraboot build was failing because it required this bump (afaict)

Itxaka (commented):
I'm not sure about this to be honest. Providing a docker file that auto builds al releases automatically like we did with enki seems like a saner approach and the pipeline already supports that. And os builder tools is provided as a docker file which makes it easy to bring binaries from other parts as we did with enki 

This creates yet another package and while it may be helpful as we can install it directly in the os builder tools, isn't the point to make auroraboot the main image and install the tools onto that image instead? So all this scripts on os builder tools would go into aurora instead?

Maybe I understood it wrongly?

jimmykarily:
> I'm not sure about this to be honest. Providing a docker file that auto builds al releases automatically like we did with enki seems like a saner approach and the pipeline already supports that. And os builder tools is provided as a docker file which makes it easy to bring binaries from other parts as we did with enki
> 
> This creates yet another package and while it may be helpful as we can install it directly in the os builder tools, isn't the point to make auroraboot the main image and install the tools onto that image instead? So all this scripts on os builder tools would go into aurora instead?
> 
> Maybe I understood it wrongly?

No you got it right. I wasn't sure about this one either. I just thought it will be good to have another way to get just the auroraboot binary since the image we will be building from now on, won't be a small one with just auroraboot but the merge of the osbuilder-tools one with the auroraboot image.
But we do provide github artifacts so if someone wants just the binary, they can get it from there.

I'll close this then.

PR #1151, opened 2024-11-25
Parse datasources in all cases except uki normal boot

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/3035

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
@@ -2,10 +2,7 @@ name: "Datasource handling"
 stages:
   rootfs.before:
     - &datasource
-      if: |
-        (grep -q "install-mode" /proc/cmdline || grep -q "nodepair.enable" /proc/cmdline ) && \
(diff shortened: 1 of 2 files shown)


packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
```diff
@@ -2,10 +2,7 @@ name: "Datasource handling"
 stages:
   rootfs.before:
     - &datasource
-      if: |
-        (grep -q "install-mode" /proc/cmdline || grep -q "nodepair.enable" /proc/cmdline ) && \
-        ([ -f /run/cos/live_mode ] || [ -f /run/cos/uki_install_mode ]) && \
-        ( [ ! -f /oem/userdata ] )
+      if: '[ ! -f /oem/userdata ] && [ ! -f /run/cos/uki_boot_mode ]'
```
jimmykarily, line 5: The removed code was a bit redundant or am I missing something? In any case, we said we want this to be enabled in all cases except when uki is booting in normal mode. So this condition should be enough, no?

Itxaka: yes, this should be enough!

Itxaka (approved):

PR #1172, opened 2024-12-11
Bump systemd-boot to 257

Itxaka:
They changed the paths for some reason


mauromorales (approved):

Itxaka:
Lets block it until we release a patch as to not bring it already. for 3.6.0

Itxaka:
nooo dont close it as it serves as reference for when we need to update as it requires changing stuff :D

PR #1178, opened 2024-12-13
Make sure /etc/machine-id is generate as early as possible in systemd

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/303
Deprecates https://github.com/mudler/yip/pull/193

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/00_rootfs.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/00_rootfs.yaml
@@ -106,6 +106,13 @@ stages:
           /var/run/cilium
           /var/snap
         PERSISTENT_STATE_BIND: "true"
+    - if: '[ -e "/sbin/systemctl" ] || [ -e "/bin/systemctl" ] || [ -e "/usr/sbin/systemctl" ] || [ -e "/usr/bin/systemctl" ]'
+      # Mask the commit service to avoid systemd messing with machine-id.
+      # https://www.freedesktop.org/software/systemd/man/latest/systemd-machine-id-commit.service.html
+      name: "Mask the commit service to avoid systemd messing with machine-id"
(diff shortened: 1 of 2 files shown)


packages/static/kairos-overlay-files/files/system/oem/00_rootfs.yaml
```diff
@@ -140,8 +140,18 @@ stages:
...
       commands:
         - cat /usr/local/etc/machine-id > /etc/machine-id
+    # If we didn't restore it already, ensure /etc/machine-id is generated
+    - if: '[ -e "/sbin/systemctl" ] || [ -e "/bin/systemctl" ] || [ -e "/usr/sbin/systemctl" ] || [ -e "/usr/bin/systemctl" ]'
+      name: "Ensure /etc/machine-id for systemd systems"
+      commands:
+        # https://www.freedesktop.org/software/systemd/man/latest/systemd-machine-id-commit.service.html
+        # https://www.freedesktop.org/software/systemd/man/latest/systemd-machine-id-setup.html
+        # Mask the commit service to avoid systemd messing with machine-id.
+        - ln -s /dev/null /etc/systemd/system/systemd-machine-id-commit.service
```
jimmykarily, line 150: Normally, since this service will kick in later, it should have no effect, so we shouldn't need to mask it but better safe than sorry?

Itxaka: Also, we should run it on the rootfs, which is the earliest stage we got I think, so we mask it asap.

jimmykarily: Good points, pushed a commit

Itxaka (approved):

mudler (approved):

PR #1196, opened 2025-01-03
Introduce kairos.pull_datasources cmdline option to force pulling them

jimmykarily:
Fixes https://github.com/kairos-io/kairos/issues/3032

If we merge this, it needs docs.

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
@@ -2,7 +2,7 @@ name: "Datasource handling"
 stages:
   rootfs.before:
     - &datasource
-      if: '[ ! -f /oem/userdata ] && [ ! -f /run/cos/uki_boot_mode ]'
+      if: '[ ! -f /oem/userdata ] && ([ ! -f /run/cos/uki_boot_mode ] || grep -q "kairos.pull_datasources" /proc/cmdline )'
       name: "Pull data from provider"
       datasource:
         providers: ["cdrom", "gcp", "openstack", "aws", "azure", "hetzner", "packet", "vultr", "digitalocean", "metaldata", "vmware", "config-drive"]
--- a/packages/static/kairos-overlay-files/collection.yaml
+++ b/packages/static/kairos-overlay-files/collection.yaml
@@ -1,4 +1,4 @@
 packages:
   - name: "kairos-overlay-files"
     category: "static"
-    version: "1.6.2"
+    version: "1.6.3"


Itxaka (approved):
Umm I'm not sure about adding it on cmdline, I hate to rely on cmdline stuff for runtime stuff as it's prone to forgetting or in the case of UKI, hardcode

packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
```diff
@@ -2,7 +2,7 @@ name: "Datasource handling"
 stages:
   rootfs.before:
     - &datasource
-      if: '[ ! -f /oem/userdata ] && [ ! -f /run/cos/uki_boot_mode ]'
+      if: '[ ! -f /oem/userdata ] && ([ ! -f /run/cos/uki_boot_mode ] || grep -q "kairos.pull_datasources" /proc/cmdline )'
```
jimmykarily, line 5: It doesn't seem to work:

```
root@localhost:/home/kairos# cat /oem/userdata
#cloud-config

stages:
  boot:
    - name: "Create a file from datasource"
      commands:
        - touch /etc/datasource-file-dimitris
root@localhost:/home/kairos# ls /etc/datasource-file-dimitris
/etc/datasource-file-dimitris
root@localhost:/home/kairos# cat /proc/cmdline
console=ttyS0 console=tty1 net.ifnames=1 rd.immucore.oemlabel=COS_OEM rd.immucore.oemtimeout=2 rd.immucore.uki selinux=0 panic=5 rd.shell=0 systemd.crash_reboot=yes install-mode
root@localhost:/home/kairos# ls  /run/cos/uki_boot_mode
/run/cos/uki_boot_mode
```

No idea why.

jimmykarily:
```
root@localhost:/home/kairos# if [ ! -f /run/cos/uki_boot_mode ] || grep -q "kairos.pull_datasources" /proc/cmdline; then echo 'ok'; fi
root@localhost:/home/kairos#
```
manually the conditions seems to work :shrug:

jimmykarily: It was my mistake, thinking the `/oem/userdata` file is re-created on every boot. I initially booted the entry that had the cmdline option and then rebooted to the entry that didn't. But the file was already there :facepalm: . If I delete the file and boot the entry with no option, the file is not created. All good.

jimmykarily:
> Umm I'm not sure about adding it on cmdline, I hate to rely on cmdline stuff for runtime stuff as it's prone to forgetting or in the case of UKI, hardcode

I would also prefer it to be in the user config instead but the overlay files don't have access to the user config, do they? The only option would be to move the condition outside this overlay file and implement this whole stage in immucore itself (where user config is available). Is there any other option I'm missing?

PR #1225, opened 2025-01-16
Add k0s Packages

wrkode:
This PR adds the creation of k0s packages for `openrc` and `systemd`.
It is explicitly targeting Kubernetes versions `1.31.3`, `1.30.7` and `1.29.11`.

Thanks for reviewing

fixes https://github.com/kairos-io/kairos/issues/3124

### DIFF
--- a/packages/k8s/k0s/build.yaml
+++ b/packages/k8s/k0s/build.yaml
@@ -1,0 +1,42 @@
+requires:
+- name: "toolchain-go-ubuntu"
+  category: "development"
+  version: ">=0"
+env:
+  # Remove any possible "+N" from the end of the version. We only bump that to
+  # avoid overwritting existing images when we want the package to be rebuilt.
+  - K0S_VERSION=v{{ regexReplaceAll "\\+\\d+$" .Values.version "" }}+k0s.{{.Values.k0s_version}}
+  {{$arch:=(default "amd64" .Values.arch)}}
+  {{ if eq $arch "arm" }}
+  - ARCH=arm64
+  {{ else }}
+  - ARCH={{ $arch }}
+  {{ end }}
+steps:
+  - c
+  # Let the installer script install service files for openrc or systemd:
+  # https://github.com/k3s-io/k3s/blob/36645e7311e9bdbbf2adb79ecd8bd68556bc86f6/install.sh#L114-L122
+  {{ if eq .Values.name "k0s-openrc" }}
+  - touch /sbin/openrc-run && chmod +x /sbin/openrc-run
+  {{ else }}
+  - touch /bin/systemctl && chmod +x /bin/systemctl
+  - mkdir -p /etc/systemd/system/
+  {{ end }}
+  - bash installer.sh
+  - bash installer.sh agent
+  - rm -rf installer.sh
+  - chmod +x /usr/local/bin/k0s
+  - upx -1 /usr/local/bin/k0s
+
+includes:
+- ^/usr/local/bin/k0s
+{{ if eq .Values.name "k0s-openrc" }}
+- ^/etc/init.d/$
+- ^/etc/init.d/k0s*
+- ^/etc/k0s/$
+- ^/etc/k0s/k0s$
+{{ else }}
+- ^/etc/systemd$
+- ^/etc/systemd/system$
+- ^/etc/systemd/system/k0s*.service$
+{{ end }}
--- a/packages/k8s/k0s/build.yaml
+++ b/packages/k8s/k0s/build.yaml
@@ -7,11 +7,12 @@ env:
 steps:
   - curl -sfL https://get.k0s.sh > installer.sh
   {{ if eq .Values.name "k0s-openrc" }}
-  - touch /sbin/openrc-run && chmod +x /sbin/openrc-run
+  # required by https://github.com/kardianos/service/blob/becf2eb62b83ed01f5e782cb8da7bb739ded2bb5/service_openrc_linux.go#L17
+  - touch /sbin/openrc-init && chmod +x /sbin/openrc-init
   {{ else }}
   -  apt-get update && apt-get install -y systemctl
   - mkdir -p /etc/systemd/system/
-  # force systemd service check by k0s installer
+  # required by https://github.com/kardianos/service/blob/becf2eb62b83ed01f5e782cb8da7bb739ded2bb5/service_systemd_linux.go#L23
   - mkdir -p /run/systemd/system
   {{ end }}
   - bash installer.sh
--- a/packages/k8s/k0s/build.yaml
+++ b/packages/k8s/k0s/build.yaml
@@ -11,7 +11,7 @@ env:
   - ARCH={{ $arch }}
   {{ end }}
 steps:
-  - c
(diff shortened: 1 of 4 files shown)


packages/k8s/k0s/build.yaml
```diff
@@ -0,0 +1,38 @@
...
+  {{ end }}
+steps:
+  - c
+  {{ if eq .Values.name "k0s-openrc" }}
+  - touch /sbin/openrc-run && chmod +x /sbin/openrc-run
+  {{ else }}
+  - touch /bin/systemctl && chmod +x /bin/systemctl
+  - mkdir -p /etc/systemd/system/
+  {{ end }}
+  - bash installer.sh
```
mauromorales, line 21: I think the installer still needs to be downloaded somewhere before this line

packages/k8s/k0s/build.yaml
```diff
@@ -0,0 +1,38 @@
+requires:
+- name: "toolchain-go-ubuntu"
+  category: "development"
+  version: ">=0"
+env:
+  - K0S_VERSION=v{{ regexReplaceAll "\\+\\d+$" .Values.version "" }}+k0s.{{.Values.k0s_version}}
```
mauromorales, line 6: This is used in k3s becuase the installer script takes that env var to decide which version of k3s to install, I wasn't able to find a `K0S_VERSION` in the [k0s docs](https://docs.k0sproject.io/v0.9.1/)

packages/k8s/k0s/collection.yaml
```diff
@@ -0,0 +1,79 @@
+packages:
+  - name: k0s-openrc
+    category: k8s
+    version: "1.31.3"
+    k3s_version: "0"
+    labels:
+      github.owner: "k0sproject"
+      github.repo: "k0s"
+      autobump.sed_script: 's/\+k0s.[0-9]//g'
```
mauromorales, line 9: This is similar to what we have in k3s, so I'm sure it will work, just keep in mind that by doing this we ignore patched versions e.g. v1.31.3+k0s.1 will not be considered because version v1.31.3+k0s.0 is already installed.

this is an internal problem that we need to solve. I thought I had a ticket for it but I cannot find it, will relate to this one instead in the meantime https://github.com/kairos-io/kairos/issues/1934

No need to worry on your side about this atm, just a not to self kind of thing

packages/k8s/k0s/build.yaml
mauromorales: Here's how we download the k3s installer for example https://github.com/kairos-io/packages/blob/f821d2c79045a738185f7ac5a23155d607d13e03/packages/k8s/k3s/build.yaml#L21

wrkode: That's right. Deleted the line by mistake

packages/k8s/k0s/build.yaml
wrkode: Please take a look here in the docs:
https://docs.k0sproject.io/v1.31.3+k0s.0/k0s-multi-node/

packages/k8s/k0s/build.yaml
wrkode: added in the last commit

packages/k8s/k0s/build.yaml
mauromorales: oh nice, my bad

mauromorales (approved):

mauromorales:
overall it lgtm, I approved the run of the pipelines and if it's green I think we can merge and start playing with it. If there are any issues we can just fix them on a new PR

packages/k8s/k0s/build.yaml
```diff
@@ -0,0 +1,40 @@
...
+  {{$arch:=(default "amd64" .Values.arch)}}
+  {{ if eq $arch "arm" }}
+  - ARCH=arm64
+  {{ else }}
```
Itxaka, line 10: do we need all this arch stuff? the installer seems to know which arch its running on so it should not be necessary to do it ourselves

packages/k8s/k0s/build.yaml
```diff
@@ -0,0 +1,40 @@
...
+  - ARCH={{ $arch }}
+  {{ end }}
+steps:
+  - curl -sfL https://get.k0s.sh > installer.sh
+  {{ if eq .Values.name "k0s-openrc" }}
+  - touch /sbin/openrc-run && chmod +x /sbin/openrc-run
+  {{ else }}
+  - touch /bin/systemctl && chmod +x /bin/systemctl
```
Itxaka, line 18: I think we need the real systemd installed so it can access the systemctl command directly, otherwise it fails to install it:

```
       #15 0.508 Error: failed to install k0s service: failed to install service: "systemctl" failed: fork/exec /usr/bin/systemctl: exec format error
```

packages/k8s/k0s/build.yaml
mauromorales: good observation

packages/k8s/k0s/build.yaml
```diff
@@ -0,0 +1,34 @@
...
+  # force systemd service check by k0s installer
+  - mkdir -p /run/systemd/system
+  {{ end }}
+  - bash installer.sh
+  - rm -rf installer.sh
+  - k0s install controller --single --debug --verbose
+  - chmod +x /usr/local/bin/k0s
+  - upx -1 /usr/local/bin/k0s
+
+includes:
```
Itxaka, line 23: do we want this or do we want to build our own package with things where we want? 

i.e. instead of just getting whatever is on /usr/loca/bin, maybe we want to construct our package with our own hierarchy no? so we can put k0s under /usr/bin for example?

see for an example systemd-boot: https://github.com/kairos-io/packages/blob/main/packages/system/systemd/build.yaml

We dont like the actual dirs and such so we create our own and copy the stuff in the places we want:

```
  - cp systemd/build/src/boot/efi/systemd-bootx64.efi /package/usr/kairos/
  - cp systemd/build/src/boot/efi/linuxx64.efi.stub /package/usr/kairos/
  - cp systemd/build/src/boot/efi/addonx64.efi.stub /package/usr/kairos/
```

That mixed with the stanza `package_dir: /package` means that the `/package` dir is gonna be the root of the package and any hierarchy under it its gonna be stored as is, so in this package when we install, we end up with the artifacts from systemd-boot under `/usr/kairos/`

Either that or we copy the k0s binary to /usr/bin before running the controller install? And then we bring that path instead of /usr/loca/bin/k0s ? So in the end image it appears under /usr/bin ?

Otherwise as you mentioned yesterday @mauromorales we will have issues with the shadowing of the dir and probably with extensions as well that overlay /usr/local/bin

mauromorales: @Itxaka I was hoping that if the luet package had it under /usr/local/bin it would not be a problem, because then i would not have to add systemd overrides, but I guess my assumption is wrong and /usr/local/bin is still not accessible. Thanks for pointing it out, I'll do the change

mauromorales: ok did the move and added the override under the static files

PR #1266, opened 2025-02-07
fix(datasource): do run stages only for new cloud config

mudler:
Follow-up of : https://github.com/kairos-io/kairos-agent/pull/679

### DIFF
--- a/packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
+++ b/packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
@@ -2,30 +2,30 @@ name: "Datasource handling"
 stages:
   rootfs.before:
     - &datasource
-      if: '[ ! -f /oem/userdata ] && ([ ! -f /run/cos/uki_boot_mode ] || grep -q "kairos.pull_datasources" /proc/cmdline )'
+      if: '[ ! -e /oem/userdata ] && ([ ! -f /run/cos/uki_boot_mode ] || grep -q "kairos.pull_datasources" /proc/cmdline )'
       name: "Pull data from provider"
       datasource:
         providers: ["cdrom", "gcp", "openstack", "aws", "azure", "hetzner", "packet", "vultr", "digitalocean", "metaldata", "vmware", "config-drive"]
-        path: "/oem"
-    - if: '[ ! -f /oem/userdata ]'
+        path: "/oem/userdata"
+    - if: '[ ! -e /oem/userdata ]'
       name: "Sentinel file for userdata"
       files:
       - path: /run/.userdata_load
   initramfs.before:
     - <<: *datasource
-    - if: '[ ! -f /oem/userdata ]'
+    - if: '[ ! -e /oem/userdata ]'
       files:
       - path: /run/.userdata_load
   # After network, if no datasource could be pulled, we stop trying
   network:
     - <<: *datasource
-    - if: '[ -f /oem/userdata ] && [ -f /run/.userdata_load ]'
+    - if: '[ -e /oem/userdata ] && [ -f /run/.userdata_load ]'
       name: "Run stages if userdata is found"
       commands:
-      - kairos-agent run-stage initramfs
-      - kairos-agent run-stage boot
+      - kairos-agent run-stage initramfs --override-cloud-init-paths /oem/userdata
+      - kairos-agent run-stage boot --override-cloud-init-paths /oem/userdata
       - rm -rf /run/.userdata_load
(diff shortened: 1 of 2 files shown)


packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
```diff
@@ -2,30 +2,30 @@ name: "Datasource handling"
...
   rootfs.before:
     - &datasource
-      if: '[ ! -f /oem/userdata ] && ([ ! -f /run/cos/uki_boot_mode ] || grep -q "kairos.pull_datasources" /proc/cmdline )'
+      if: '[ ! -e /oem/userdata ] && ([ ! -f /run/cos/uki_boot_mode ] || grep -q "kairos.pull_datasources" /proc/cmdline )'
       name: "Pull data from provider"
       datasource:
         providers: ["cdrom", "gcp", "openstack", "aws", "azure", "hetzner", "packet", "vultr", "digitalocean", "metaldata", "vmware", "config-drive"]
-        path: "/oem"
-    - if: '[ ! -f /oem/userdata ]'
+        path: "/oem/userdata"
```
Itxaka, line 9: you sure this is correct?

Itxaka: no this is not, it will use this as the basepath and add userdata at the end if Im reading this correctly: https://github.com/mudler/yip/blob/master/pkg/plugins/datasource.go#L246

Itxaka (changes requested):
datasource path is wrong

mudler:
> datasource path is wrong

this is intentional. With this change actually we are doing two things:

1) move the datasource from `/oem` to `/oem/userdata`: that is, instead of dropping the userdata files directly into `/oem`, to create a folder with these files first
2) run the stages against the new folder so it is isolated from running all the cloud init config paths

packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
mudler: yup this is wanted. I want to create files in `/oem/userdata` (directory)

Itxaka:
> > datasource path is wrong
> 
> this is intentional. With this change actually we are doing two things:
> 
>     1. move the datasource write path from `/oem` to `/oem/userdata`: that is, instead of dropping the userdata files directly into `/oem`, to create a folder with these files (e.g. currently `/oem/userdata` and `/oem/userdata.yaml`)
> 
>     2. run the stages against the new folder so it is isolated from running all the cloud init config paths

This will create a /oem/userdata/userdata file on that second pass, which may create issues? If userdata already run, it wont be able to create that as a dir? I would choose another base path in that case. In fact, I would not even touch that xD

packages/static/kairos-overlay-files/files/system/oem/00_datasource.yaml
mudler: Current:
```
/oem/userdata /oem/userdata.yaml
```

wanted:

```
/oem/userdata/userdata /oem/userdata/userdata.yaml

```

So I can point `/oem/userdata` as a cloud init path

Itxaka (approved):

mudler:
> > > datasource path is wrong
> > 
> > 
> > this is intentional. With this change actually we are doing two things:
> > ```
> > 1. move the datasource write path from `/oem` to `/oem/userdata`: that is, instead of dropping the userdata files directly into `/oem`, to create a folder with these files (e.g. currently `/oem/userdata` and `/oem/userdata.yaml`)
> > 
> > 2. run the stages against the new folder so it is isolated from running all the cloud init config paths
> > ```
> 
> This will create a /oem/userdata/userdata file on that second pass, which may create issues? If userdata already run, it wont be able to create that as a dir? I would choose another base path in that case. In fact, I would not even touch that xD

Which second pass?

Currently this is how the datasource overlays things into oem:

```
/oem/userdata /oem/userdata.yaml
```

With this change, I'm now just specifying the userdata to be created as a folder in `/oem`, that is

```
/oem/userdata/userdata /oem/userdata/userdata.yaml

```

I don't move it out of oem for two reasons:

- we want to keep the same behavior (persist what was pulled during userdata, because it might come from removable devices e.g. USB)
- I need a folder to point the cloud init executor against. I could also give directly it a file, however, if datasource is not a yaml file it won't write it - so to avoid this ambiguity I let the cloud config scanner to take it over by giving it a path (in this way the scanner will identify valid cloud config files, and I don't have to do it there)

PR #1293, opened 2025-02-28
add RPI 5 firmware

zreigz:
This PR add RPI 5 firmware


Itxaka:
We usually get the artifacts directly from the source or a distro, so they are verifiable. 
Seems like if the artifacts come from home assistant, they also use the upstream firmware files directly:

https://github.com/home-assistant/operating-system/blob/dev/scripts/update-firmware-rpi.sh

So we should be able to just add a package that has the full rpi firmware for all model directly from upstream. Nice!

Let me have a quick test at this

PR #1308, opened 2025-03-07
Rework the toolchain

Itxaka:
- rename it to toolchain as thats what is mainly used for
 - drop alpine as it was never used
 - use ubuntu 20.04 as base which has an older glibc
 - add bumping to versions automatically
 - Get go from the upstream source directly
 - Add default deps to it, like gcc and git
 - make everything use the newer toolchain-go
 - drop uneeded toolchain-ubuntu
 - Move packages around
 - Drop all k8s packages as they are no longer used
 - Add as much deps as possible part of the toolchain image
 - update toolchain nodejs and npn to stable LTS version

Part of https://github.com/kairos-io/kairos/issues/3231


packages/k8s/kubeadm/build.yaml
mauromorales, line 1: are we sure these are not in use anymore? I see the provider's are still maintained https://github.com/kairos-io/provider-kubeadm

Itxaka: Oh I thought they provided their own packages for that. Weird that the provider is off but we still have the package here... Will restore, good catch!

packages/system/kairos-agent/build.yaml
```diff
@@ -1,20 +1,22 @@
...
 requires:
-  - name: "toolchain-go-ubuntu"
+  - name: "toolchain-go"
     category: "development"
     version: ">=0"
 {{end}}
 prelude:
-  - apt-get update && apt-get install -y npm gcc
+  - curl -fsSL https://deb.nodesource.com/setup_22.x | bash - # Install Node.js 22.x via nodesource to get the stable supported version
+  - apt-get update && apt-get install -y nodejs
```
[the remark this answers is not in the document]
Itxaka: Yes indeed, will do and also rename it to toolchain-base or something like that so it's not go only

packages/k8s/kubeadm/build.yaml
Itxaka: Actually they seem to get their own binaries: https://github.com/kairos-io/provider-kubeadm/blob/main/Earthfile#L99

mauromorales (approved):
lgtm

PR #1309, opened 2025-03-10
Move rockchip package to different category

Itxaka:
Make it in line with others. Also dont be opinionated on where the needed files are set, just drop them in the root package dir like we do with others


packages/uboot/rockchip/build.yaml
```diff
@@ -0,0 +1,14 @@
+requires:
+  - name: "toolchain"
+    category: "development"
+    version: ">=0"
+
+package_dir: /package
+
+prelude:
+- wget http://http.us.debian.org/debian/pool/main/u/u-boot/u-boot-rockchip_2025.01-2_arm64.deb -O u-boot.deb
```
mauromorales, line 9: shouldn't there be a variable with the version here?

mauromorales (approved):
made a comment but I see it's also how it was done before so either way lgtm

packages/uboot/rockchip/build.yaml
Itxaka: Ideally we should move to build it ourselves if possible, so we got the most vanilla binaries out of this, but this was never updated so I just kept like that. Not even sure anyone is using it or even if we have one of these anymore lol

PR #1310, opened 2025-03-10
Patch our shipped systemd-boot to support poweroff after enroll

Itxaka:
Upstream patch to systemd >257 at https://github.com/systemd/systemd/pull/36684

I would say this is blocked until we know more about the expected workflow, but this also helps if you just want to enroll and ship machines, you can add the iso to a node and just wait until it shut downs, as that would mark that the enrollment went ok, so you then ship it. And you can use the same ISO that is going to be used to install it afterwards, so no need to generate new ones and such.

Still, lets wait maybe? Although I think its a great addition


mauromorales (approved):
lgtm

mudler (approved):

PR #1323, opened 2025-03-14
Restore k3s packages for patch release

Itxaka:


jimmykarily (approved):
Is this the right way?

PR #1378, opened 2025-04-17
Bump k3s versions to latest patch

jimmykarily:


Itxaka (approved):
Is this useful? On 3.4 we use the upstream packages directly so this is moot at this point no?

jimmykarily:
> Is this useful? On 3.4 we use the upstream packages directly so this is moot at this point no?

true: https://github.com/kairos-io/kairos/blob/823b4906e0c7df8168d5dff9631619a8e2f3037e/.github/workflows/release.yaml#L96
forgot about that. We should decide if we keep these packages then.

PR #1386, opened 2025-04-25
Check software version too when deciding if to upgrade

romario-07:
For older versions, this check was enough: KAIROS_VERSION="v3.3.6-v1.32.2-k3s1" to decide if upgrade is needed or not, because the version included k3s version too.

In the latest version, k3s version is not included and this check is not enough: KAIROS_VERSION="v3.4.0". 
KAIROS_SOFTWARE_VERSION="1.32.3+k3s1" has to be checked too, in case the same Kairos version is used, but different k3s version.

This should fix the case when upgrading from "12-standard-amd64-generic-v3.4.0-k3s1.30.11-k3s1" to "12-standard-amd64-generic-v3.4.0-k3s1.31.7-k3s1". Right now, it won't upgrade and it will report that the system is up to date.

This is how /etc/kairos-release looks like in the latest version:
image

KAIROS_SOFTWARE_VERSION is the only piece of information regarding k3s version.

### DIFF
--- a/packages/system/suc-upgrade/suc-upgrade.sh
+++ b/packages/system/suc-upgrade/suc-upgrade.sh
@@ -8,18 +8,18 @@ echo "SUC_VERSION: $SUC_VERSION"
 if [ "$FORCE" != "true" ]; then
     if [ -f "/etc/kairos-release" ]; then
       # shellcheck disable=SC1091
-      UPDATE_VERSION=$(source /etc/kairos-release && echo "${KAIROS_VERSION}")
+      UPDATE_VERSION=$(source /etc/kairos-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}")
     else
       # shellcheck disable=SC1091
-      UPDATE_VERSION=$(source /etc/os-release && echo "${KAIROS_VERSION}")
+      UPDATE_VERSION=$(source /etc/os-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}"")
     fi
 
     if [ -f "${HOST_DIR}/etc/kairos-release" ]; then
       # shellcheck disable=SC1091
-      CURRENT_VERSION=$(source "${HOST_DIR}"/etc/kairos-release && echo "${KAIROS_VERSION}")
+      CURRENT_VERSION=$(source "${HOST_DIR}"/etc/kairos-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}"")
     else
       # shellcheck disable=SC1091
-      CURRENT_VERSION=$(source "${HOST_DIR}"/etc/os-release && echo "${KAIROS_VERSION}")
+      CURRENT_VERSION=$(source "${HOST_DIR}"/etc/os-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}"")
     fi
 
     if [ "$CURRENT_VERSION" == "$UPDATE_VERSION" ]; then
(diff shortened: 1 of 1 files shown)


packages/system/suc-upgrade/suc-upgrade.sh
```diff
@@ -8,18 +8,18 @@ echo "SUC_VERSION: $SUC_VERSION"
 if [ "$FORCE" != "true" ]; then
     if [ -f "/etc/kairos-release" ]; then
       # shellcheck disable=SC1091
-      UPDATE_VERSION=$(source /etc/kairos-release && echo "${KAIROS_VERSION}")
+      UPDATE_VERSION=$(source /etc/kairos-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}")
     else
       # shellcheck disable=SC1091
-      UPDATE_VERSION=$(source /etc/os-release && echo "${KAIROS_VERSION}")
+      UPDATE_VERSION=$(source /etc/os-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}"")
```
Itxaka, line 14:
```suggestion
      UPDATE_VERSION=$(source /etc/os-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}")
```

I think there is an extra quote here?

packages/system/suc-upgrade/suc-upgrade.sh
```diff
@@ -8,18 +8,18 @@ echo "SUC_VERSION: $SUC_VERSION"
...
     fi
 
     if [ -f "${HOST_DIR}/etc/kairos-release" ]; then
       # shellcheck disable=SC1091
-      CURRENT_VERSION=$(source "${HOST_DIR}"/etc/kairos-release && echo "${KAIROS_VERSION}")
+      CURRENT_VERSION=$(source "${HOST_DIR}"/etc/kairos-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}"")
```
Itxaka, line 19: same here

Itxaka (approved):
its missing the package version bump so it gets built!

under the definition.yaml the version should be bumped, otherwise the package wont be built!

packages/system/suc-upgrade/suc-upgrade.sh
```diff
@@ -8,18 +8,18 @@ echo "SUC_VERSION: $SUC_VERSION"
 if [ "$FORCE" != "true" ]; then
     if [ -f "/etc/kairos-release" ]; then
       # shellcheck disable=SC1091
-      UPDATE_VERSION=$(source /etc/kairos-release && echo "${KAIROS_VERSION}")
+      UPDATE_VERSION=$(source /etc/kairos-release && echo "KAIROS_VERSION=${KAIROS_VERSION};KAIROS_SOFTWARE_VERSION=${KAIROS_SOFTWARE_VERSION}")
```
Itxaka, line 11: maybe we could just merge both keys so it behaves as before. You are saying that before it would be VERSION+K3SVERSION so maybe a simpler thing would be to do 

```
UPDATE_VERSION=$(source /etc/kairos-release && echo "${KAIROS_VERSION}${KAIROS_SOFTWARE_VERSION}")
```

thoughs?

Itxaka:
umm in fact, we should even go further and construct the "final" version by merging the kairos versions + software provider (k3s, k0s) and the provider version (1.32.3+k3s1)

As now we can have a different provider of k8s, we dont want to upgrade from a k3s to a k0s and so on

mudler:
> umm in fact, we should even go further and construct the "final" version by merging the kairos versions + software provider (k3s, k0s) and the provider version (1.32.3+k3s1)
> 
> As now we can have a different provider of k8s, we dont want to upgrade from a k3s to a k0s and so on

yes agreed, we should take into account of all the involved pieces, this means to also use `KAIROS_SOFTWARE_VERSION_PREFIX` too (didn't we had a specific one for the kubernetes distro? it would be more intuitive in this case)

Itxaka:
> > umm in fact, we should even go further and construct the "final" version by merging the kairos versions + software provider (k3s, k0s) and the provider version (1.32.3+k3s1)
> > As now we can have a different provider of k8s, we dont want to upgrade from a k3s to a k0s and so on
> 
> yes agreed, we should take into account of all the involved pieces, this means to also use `KAIROS_SOFTWARE_VERSION_PREFIX` too (didn't we had a specific one for the kubernetes distro? it would be more intuitive in this case)

the prefix is the k8s provider :)

Terible naming yes, we need to transition that into a more understandable key

romario-07:
I see there is another pr fixing the issue (https://github.com/kairos-io/packages/pull/1389). Should this one be closed?

Itxaka:
> I see there is another pr fixing this issue (https://github.com/kairos-io/packages/pull/1389). Should this one be closed?

Yes, sorry, I forgot to ping you in the PR as we were testing this in place!

PR #1461, opened 2025-09-26
Improve tiny kernel

Itxaka:
This builds the minimum size kernel to support systemd, services, cgroups, kairos stuff and so on

This generates a tiny kernel with no modules and eveyrhting needed bundled on it, for example:

 - for mounting loop devices and squash, modules are in the kernel directly
 - support for cgroups
 - iso9660 support
 - qemu virtio drivers

This makes it usable with a distro for testing while keeping the kernel under 4Mb.

The idea would be to then create a separate package with the same config but other things as modules. Maybe even several packages, like modules-blk, modules-gpu, modules-snd and so on and put those separated in packages so its not a huge 300Mb modules bundled


mauromorales (approved):
didn't really check all those options 😅 just some random ones that resonated with what you wrote on the description :P

Itxaka:
> didn't really check all those options 😅 just some random ones that resonated with what you wrote on the description :P 

Not valid, you need to check each one against the kernel docs

Crap, I didn't wanted this to be merged yet

